PR #3, opened 2016-03-12
Add docker as a platform option

alexcu:
# Problem

Up until now, Doubtfire has been hard to set up especially for novice users. There should be an easy way to get Doubtfire up and running without too much hassle, even for demonstration purposes.

# What's Changed?

This pull request adds Docker config to the api so that it is possible to run it via Docker.

To do so, use the extensive `./docker.sh` script.


alexcu:
As docker is now used, the [`config/add-vagrant`]( https://github.com/doubtfire-lms/doubtfire-api/tree/config/add-vagrant) branch is (in addition to stale) redundant and should be deleted.

PR #6, opened 2016-03-25
Enhance/add time submitted and assessed

Reubsinit:
This all works on my end. Let me know what you guys think @macite @alexcu. Was I supposed to omit the changes to schema so you could run the `rake db:migrate` with the script I added?

### DIFF
--- a/app/api/task_definitions.rb
+++ b/app/api/task_definitions.rb
@@ -272,7 +272,7 @@ module Api
       unit.student_tasks.
         joins(:project).
         joins(:task_status).
-        select("projects.tutorial_id as tutorial_id", "project_id", "tasks.id as id", "task_definition_id", "task_statuses.name as status_name", "completion_date", "times_assessed").
+        select("projects.tutorial_id as tutorial_id", "project_id", "tasks.id as id", "task_definition_id", "task_statuses.name as status_name", "completion_date", "times_assessed", "times_submitted").
         where("task_definition_id = :id", id: params[:task_def_id]).
         map { |t|
         {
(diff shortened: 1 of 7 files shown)


Reubsinit:
Also, I was thinking, what about students who have already submitted things before this change?

So I was thinking for those who have already submitted:
```if times_assessed > 0 and times_submitted == 0 then times_submitted = times_assessed - 1```

macite:
In submits can you change it to only increment the number of submissions if the task is not currently in the ready to mark status.

macite:
Also need to include an update to the latex task cover sheet. Add time so submission, times assessed and submitted, and students id. Could also add a list of comments if we can get it to fit.

Reubsinit:
@macite Was going to wait to ask you about latex because the dev version doesn't seem to render any pdfs on the tasks I tried to submit through. So I can't actually see my changes.

Reubsinit:
and also, do you want something like:
```def submit(submit_date = Time.zone.now)
      if self.task_status != TaskStatus.ready_to_mark
           self.times_submitted += 1
      self.task_status      = TaskStatus.ready_to_mark
      self.submission_date  = submit_date```

or 

```def submit(submit_date = Time.zone.now)
      if self.task_status != TaskStatus.ready_to_mark
           self.times_submitted += 1
           self.task_status      = TaskStatus.ready_to_mark
      self.submission_date  = submit_date```

Reubsinit:
Also, in that conditional, do you want to assign the task status to complete as well as incrementing the submission count? Because if the task status is not complete upon submitting, then it should only really be complete then.

macite:
You can generate the PDFs using rake submission:generate_pdfs. You can then download the PDFs but not view them with the local host server.

cheers,
Andrew

Reubsinit:
@macite Should be good to go.

macite:
@Reubsinit looking good. I'll merge it when I'm back during the week.

PR #8, opened 2016-04-06
Test/add minitest framework

Reubsinit:
This merge will allow for people to start developing tests using the `mini-test` framework.


db/migrate/20130613070051_rename_project_template_id_to_unit_id.rb
```diff
@@ -5,8 +5,8 @@ def change
     rename_column :project_convenors, :project_template_id, :unit_id
     rename_column :projects, :project_template_id, :unit_id
 
-    rename_index :projects, "index_projects_on_project_template_id", "index_projects_on_unit_id"
-    rename_index :task_templates, "index_task_templates_on_project_template_id", "index_task_templates_on_unit_id"
-    rename_index :teams, "index_teams_on_project_template_id", "index_teams_on_unit_id"
+    # rename_index :projects, "index_projects_on_project_template_id", "index_projects_on_unit_id"
```
macite: Why does this file need to change? These are old migrations.

db/schema.rb
```diff
@@ -119,6 +119,13 @@
   add_index "plagiarism_match_links", ["other_task_id"], name: "index_plagiarism_match_links_on_other_task_id", using: :btree
   add_index "plagiarism_match_links", ["task_id"], name: "index_plagiarism_match_links_on_task_id", using: :btree
 
+  create_table "project_convenors", force: true do |t|
```
macite: Something seems wrong here... the database schema shouldn't have changed for this.

spec/factories/group_sets_factory.rb
```diff
@@ -1,8 +0,0 @@
-# Read about factories at https://github.com/thoughtbot/factory_girl
```
macite: It would be good to keep the old spec tests until they are switched to minitest.

PR #10, opened 2016-04-09
TEST: Add test for API end point auth

Reubsinit:
This sees the inclusion of endpoint testing for `auth` for `post`, `put` and `delete`

### DIFF
--- a/test/api/auth_test.rb
+++ b/test/api/auth_test.rb
@@ -28,11 +28,11 @@ class AuthTest < MiniTest::Test
     expected_auth = User.first
 
     # Check to see if the username matches what was expected
-    assert_equal actual_auth['user']['username'], User.first.username
+    assert_equal actual_auth['user']['username'], expected_auth.username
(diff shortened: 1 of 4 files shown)


test/api/auth_test.rb
```diff
@@ -0,0 +1,36 @@
...
+class AuthTest < MiniTest::Test
+  include Rack::Test::Methods
+
+  def app
+    Rails.application
+  end
+
+  def test_auth_post
+    # Get response back for logging in with username 'acain' password 'password'
+    post '/api/auth.json', '{"username":"acain", "password":"password"}', "CONTENT_TYPE" => 'application/json'
```
alexcu: @Reubsinit Awesome! Where did you end up finding how to do this? Or was it literally just playing with the library code?

jakerenzella: There was one little blog post but most of it was Reuben figuring it out :)

We are writing up good documentation for it.

PR #12, opened 2016-04-11
Wrap `Terminator` in a helper called `timeout_helper`

alexcu:
Provides a more descriptive way of using `Terminator`'s that will rescue with an optional message if timeout occurs. This means more descriptive messages can be responded back to users (was getting "Invalid document provided" now we can get "Server couldn't validate pdf due to timeout")


app/helpers/file_helper.rb
```diff
@@ -187,11 +187,11 @@ def compress_image(path)
...
       # puts exec
 
-      # try with convert
-      did_compress = false
-      Terminator.terminate 40 do
-        did_compress = system exec
-      end
+    # try with ghostscript
+    did_compress = false
+    try_within 120, "compressing image" do
```
macite: Could you change these to the newer shorter times... 2 minutes ended up being too much here. 40 sec now... though I think it will be less later.

alexcu: Sure thing 😊 sorry must've missed the new values

PR #14, opened 2016-04-23
ENHANCE: Change student import to use col labels

lachlanwest:
Order of columns was important before this change. Now, as long
as the correct columns are present, the import will succeed.


macite:
The first column should be `unit_code`. Have it so that you can export and then import that file without issue. Otherwise good.

app/models/unit.rb
```diff
@@ -363,13 +363,23 @@ def unenrol_users_from_csv(file)
 
         project_participant = project_participant.first
 
-        user_project = projects.where(user_id: project_participant.id).first
```
macite: Looks like this change has not gone across. UnitRole is for staff only, so update to use this in the new change.

app/models/unit.rb
```diff
@@ -363,13 +363,23 @@ def unenrol_users_from_csv(file)
...
+        user_project = UnitRole.joins(project: :unit).where(
+            user_id: project_participant.id,
+            projects: {unit_id: id}
+          )
 
-        unless user_project
```
macite: Also this...

app/models/unit.rb
```diff
@@ -1055,7 +1065,7 @@ def path_to_task_pdf(task_def)
   def tasks_awaiting_feedback
     student_tasks.
       joins(:task_status).
-      select("project_id", "tasks.id as id", "task_definition_id", "projects.tutorial_id as tutorial_id", "task_statuses.name as status_name", "completion_date", "times_assessed", "submission_date", "portfolio_evidence").
+      select("project_id", "tasks.id as id", "task_definition_id", "projects.tutorial_id as tutorial_id", "task_statuses.name as status_name", "completion_date", "times_assessed", "portfolio_evidence", "submission_date", "times_submitted").
```
macite: times submitted has been removed... So this shouldn't have changed.

app/models/unit.rb
```diff
@@ -1068,6 +1078,7 @@ def tasks_awaiting_feedback
           status: TaskStatus.status_key_for_name(t.status_name),
           completion_date: t.completion_date,
           submission_date: t.submission_date,
+          times_submitted: t.times_submitted,
```
macite: Same

app/models/unit.rb
```diff
@@ -363,13 +363,23 @@ def unenrol_users_from_csv(file)
...
         project_participant = project_participant.first
 
-        user_project = projects.where(user_id: project_participant.id).first
+        user_project = UnitRole.joins(project: :unit).where(
+            user_id: project_participant.id,
+            projects: {unit_id: id}
+          )
 
-        unless user_project
+        if not user_project
```
macite: Does this run?

PR #22, opened 2016-07-08
Add minitest framework to begin supporting TDD approach for API changes

alexcu:
@jakerenzella and I discussed switching to a TDD approach to unit testing for API changes. We suggested to merge this branch into `develop`; therefore new features (such as the ticketing system) can be branched off `develop` and a TDD approach can be adopted using these additions as a template/guide.

Additional tests that are yet to be written for existing APIs can then be branched off developed accordingly, e.g., `test/api-submission`, `test/api-group-sets` etc.

What do you think @macite?


macite:
Agreed. Now that the issues are sorted with this I am very keen to move to TDD... including using this when fixing other issues as well etc.

PR #23, opened 2016-07-08
Improve JSON for unit testing and minor additions to setting up first-time unit tests

alexcu:
This PR adds some helpers such as `post_json` and `put_json`, where:

```ruby
post '/api/auth.json', data_to_post.to_json, "CONTENT_TYPE" => 'application/json'
```

becomes

```ruby
post_json '/api/auth.json', data_to_post
```

Also removes the ticket test which shouldn't have been added yet.

In addition, with the transactions working, we only need to set up the database once for testing—if the `doubtfire_dev_test` db does _not_ exist, then it will setup and populate it once. There was no reference to this in the docs.

A testing section has also been included in the README.md

Forgot to include these in the PR you approved earlier @macite 😉


test/helpers/json_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    #
+    # Assert that a JSON response matches the model and keys provided
+    #
+    def assert_json_matches_model(response_json, model, keys)
+      keys.each { |k| assert response_json.has_key?(k), "Response has key #{k}"}
+      keys.each { |k| assert_equal model[k], response_json[k], "Values for key #{k} match" }
+    end
+
+    module_function :assert_json_matches_model
+    module_function :post_json
```
macite: @alexcu Should put_json also be exported?

test/test_helper.rb
```diff
@@ -1,18 +1,31 @@
...
-# Require help with all tests
-require 'helpers/auth_helper'
-require 'helpers/assert_helper'
+# Check if we're connected to the test DB
+begin
+  ActiveRecord::Base.connection
+rescue ActiveRecord::NoDatabaseError
+  # No database... try setting up
+  puts "No test database has been setup! Setting first-time up..."
+  system "RAILS_ENV=test rake db:setup db:init_test_data"
```
macite: @alexcu Have a quick check -- but I recall finding a way of running rake tasks from rails without needing a system call... if we can change this that would be better.

alexcu: Likewise. It was originally there to begin with but I think Jake and I had some issues trying to get that working a while back. Let me give it another go :)

test/helpers/json_helper.rb
alexcu: Whoops! Yes that's right

alexcu:
@macite Those fixes have been made. I have also re-introduced the `DatabasePopulator` helper class. Eventually it would be nice to make all of `db:populate` and `db:init` reference this one helper class to reduce the number of duplicated code.

macite:
Yeah looks good!

PR #25, opened 2016-07-23
Fix minitest framework and add additional helpers

alexcu:
@jakerenzella and myself worked on fixing the test framework to ensure all tests use minitest and that transactions are used on our [new/helpdesk-ticketing](https://github.com/final-year-project/doubtfire-api/tree/new/helpdesk-ticketing) branch. This pull request cherry-picks those commits into this branch so that it can be implemented in [develop](https://github.com/doubtfire-api/doubtfire-api/tree/develop)

PR #26, opened 2016-07-23
Refactor Gemfile dependencies

alexcu:
This PR refactors the Gemfile to reduce unnessasary installations of gems where possible. It also updates the `:test` group now that the framework is ready to go (see #25). When #25 is merged, there will be some conflicts since this PR does some minor refactoring on the Gemfile also. I will fix these conflicts, but proposing these changes so that @macite can review what has been modified.

In addition, the documentation now removes the `test` group from the `--without` bundle switch, since testing is to be encouraged. Hence, we want bundler to install the `test` group.

I'm working on a `demo` branch Docker image (as discussed with you @macite) [here](https://github.com/final-year-project/doubtfire-api/tree/demo) which obviously won't need the `test` group—this will speedup the time for the Docker image to build and will be helpful in situations where we don't need tests to be installed.


macite:
Is this needed any more?

alexcu:
@macite Yes, this PR makes some changes to the Gemfile grouping so that we don't need mysql2 on the test environment (amongst others). When installing we should now use:

```
$ bundle install --without production replica
```

as we want `test` gems to be included for unit testing.

PR #44, opened 2016-12-21
Stub new task inbox endpoint

alexcu:
Adds endpoint for task inbox. Stubs data as applicable. Added `TODO: JAKE - <msg>` for areas that need real data for @jakerenzella to do as discussed with @macite.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -205,7 +205,7 @@ class Task < ActiveRecord::Base
 
   # Returns the number of new comments for a user
   def new_comments_for_user(user)
-    comments.map { |c| c.new_for_user?(user) }.compact.length
+    comments.where(recipient: user, is_new: true).count
   end
 
   delegate :due_date, to: :task_definition
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -188,9 +188,9 @@ class Task < ActiveRecord::Base
   # submission date or latest student comment -- whichever is newer
   def action_date
     return nil if last_student_comment.nil? || submission_date.nil?
-    return last_student_comment if !last_student_comment.nil? && submission_date.nil?
-    return submission_date      if !submission_date.nil? && last_student_comment.nil?
-    last_student_comment > submission_date ? last_student_comment : submission_date
+    return last_student_comment.created_at if !last_student_comment.nil? && submission_date.nil?
+    return submission_date.created_at      if !submission_date.nil? && last_student_comment.nil?
+    last_student_comment.created_at > submission_date ? last_student_comment.created_at : submission_date
   end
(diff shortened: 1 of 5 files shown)


app/models/task.rb
```diff
@@ -184,6 +184,30 @@ def days_overdue
     (project.reference_date - task_definition.target_date).to_i / 1.day
   end
 
+  # Action date defines the last time a task has been "actioned", either the
+  # submission date or latest student comment -- whichever is newer
+  def action_date
+    return nil if last_student_comment.nil? || submission_date.nil?
+    return last_student_comment if !last_student_comment.nil? && submission_date.nil?
```
macite: Shouldn't this return the date of that comment?

app/models/task.rb
```diff
@@ -184,6 +184,30 @@ def days_overdue
...
+    return submission_date      if !submission_date.nil? && last_student_comment.nil?
+    last_student_comment > submission_date ? last_student_comment : submission_date
```
macite: Similar here?

app/models/task.rb
```diff
@@ -184,6 +184,30 @@ def days_overdue
...
+  end
+
+  # Returns the last tutor comment for this task
+  def last_tutor_comment
+    comments.where(user: project.tutorial.tutor.user).order(:created_at).last
+  end
+
+  # Returns the number of new comments for a user
+  def new_comments_for_user(user)
+    comments.map { |c| c.new_for_user?(user) }.compact.length
```
macite: I suggest doing this with a query. Use where to limit tasks to those where recipient_id is user.id, and then count to get the count of these.

app/serializers/task_serializer.rb
```diff
@@ -3,20 +3,26 @@ class ShallowTaskSerializer < ActiveModel::Serializer
...
+             :quality_pts,
+             :num_new_comments
 
   def status
     TaskStatus.status_key_for_name(object.status_name)
   end
+
+  def num_new_comments
+    # TODO: JAKE - Stub using user first -- need to get actual current user
+    object.new_comments_for_user(User.first)
```
macite: I suggest delete this method from Task, and have this return a random number...

app/models/task.rb
alexcu: Fixed in 8a3e1e6

app/models/task.rb
alexcu: Ok

alexcu:
@jakerenzella Did you end up working from this fork for the new endpoints?

alexcu:
If so I can close this and you can send through a PR for that instead.

PR #48, opened 2017-01-13
Feature/add student_id to user

jakerenzella:
This PR will add the migration, model validations and support for User.student_id

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161208055326) do
+ActiveRecord::Schema.define(version: 20170112231900) do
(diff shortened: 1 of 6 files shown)


alexcu:
@jakerenzella Can we check that `student_id` is only added to those whose system role is `student`?

db/schema.rb
```diff
@@ -387,6 +387,7 @@
     t.boolean  "opt_in_to_research"
     t.boolean  "has_run_first_time_setup",                    default: false
     t.string   "login_id"
+    t.string   "student_id"
   end
 
   add_index "users", ["authentication_token"], name: "index_users_on_authentication_token", unique: true, using: :btree
```
alexcu: Should we add an index for `student_id`? Just a thought if we're going to be using it a lot.

PR #50, opened 2017-01-17
FIX: Only use password field in User if AAF is not enabled

jakerenzella:
This PR will ensure that the database populator is successful when AAF is enabled by not assigning a password to the password field on a User (which does not exist if AAF is enabled)

### DIFF
--- a/lib/helpers/find_or_create_students.rb
+++ b/lib/helpers/find_or_create_students.rb
@@ -5,16 +5,27 @@ def find_or_create_student(username)
   user_created = nil
   using_cache = !!@user_cache
   if !using_cache || !@user_cache.key?(username)
-    profile = {
-      first_name:             Faker::Name.first_name,
-      last_name:              Faker::Name.last_name,
-      nickname:               username,
-      role_id:                Role.student_id,
-      email:                  "#{username}@doubtfire.com",
-      username:               username,
-      password:               'password',
-      password_confirmation:  'password'
-    }
+    if AuthenticationHelpers.aaf_auth?
+      profile = {
+        first_name:             Faker::Name.first_name,
+        last_name:              Faker::Name.last_name,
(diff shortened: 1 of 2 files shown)


lib/helpers/find_or_create_students.rb
```diff
@@ -5,16 +5,27 @@ def find_or_create_student(username)
...
+      }
+    else
+      profile = {
+        first_name:             Faker::Name.first_name,
+        last_name:              Faker::Name.last_name,
+        nickname:               username,
+        role_id:                Role.student_id,
+        email:                  "#{username}@doubtfire.com",
+        username:               username,
+        password:               'password',
```
alexcu: Rather than duplicate the hash, why not define it and only add `password` and `password_confirmation` unless `AuthenticationHelpers.aaf_auth?`?

```ruby
profile = { ... }
unless AuthenticationHelpers.aaf_auth?
  profile[:password] = profile[:password_confirmation] = 'password'
end
```

jakerenzella:
Ready to merge 👍

lib/helpers/find_or_create_students.rb
```diff
@@ -5,26 +5,17 @@ def find_or_create_student(username)
...
-        nickname:               username,
-        role_id:                Role.student_id,
-        email:                  "#{username}@doubtfire.com",
-        username:               username,
-        password:               'password',
-        password_confirmation:  'password'
-      }
+        username:               username
+    }
+    if !AuthenticationHelpers.aaf_auth?
```
alexcu: Rubocop will turn/suggest to turn this this into `unless`

jakerenzella: Forgot I disabled rubocop temporarily and forgot to re-enable it, will fix

PR #51, opened 2017-01-17
Fix issues with groups and assign number to groups for reference

alexcu:
Adds a `group_number` concept so that a group may have a name and a consistent number.

Also fixes a number of issues:
- remove group member if not part of group fixed
- prevent switching a group's tutorial if the group set does not allow groups to be part of multiple tutorials
- ensures students from the _right_ tutorial can only be added to a tutorial (given non-mixed tutorial groupset)
- prevent member from being added to group twice

### DIFF
--- a/app/api/group_sets.rb
+++ b/app/api/group_sets.rb
@@ -171,7 +171,8 @@ module Api
           group_params[:name] = "Group #{id}"
         end
       end
-      grp = Group.create(name: group_params[:name], group_set: group_set, tutorial: tutorial)
+      num = group_set.groups.last.number + 1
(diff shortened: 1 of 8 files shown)


app/api/group_sets.rb
```diff
@@ -287,6 +294,14 @@ class GroupSets < Grape::API
         error!({ error: 'Not authorised to manage this student' }, 403)
       end
 
+      if gs.keep_groups_in_same_class && prj.tutorial != grp.tutorial
+        error!({ error: "tudents from the tutorial '#{grp.tutorial.abbreviation}' can only be added to this group." }, 403)
```
macite: Missing "s" on students

alexcu: Fixed

PR #68, opened 2017-07-04
Fix installation script for Sierra compatibility

alexcu:
Found several issues when installing on Sierra (had to install on my DSTIL Mac). Other modifications such as removing `replica` installation with `staging`.

`Gemfile.lock` was substantially updated.


alexcu:
Outdated gemfile. Will resubmit once updated.

PR #70, opened 2017-10-23
Add optional argument to download PDF as attachment

alexcu:
This PR fixes issues in https://github.com/doubtfire-lms/doubtfire-web/pull/81.

The native `object` in Chrome and Firefox must receive a MIME type `application/pdf` **without** the `Content-Disposition` header set as `attachment`. Safari seems to ignore this, which is why it was rendering in Safari.

### DIFF
--- a/app/api/submission/portfolio_api.rb
+++ b/app/api/submission/portfolio_api.rb
@@ -65,6 +65,9 @@ module Api
       end
 
       desc 'Retrieve portfolio for project with the given id'
+      params do
+        optional :as_attachment, type: Boolean, desc: 'Whether or not to download file as attachment. Default is false.'
+      end
       get '/submission/project/:id/portfolio' do
         project = Project.find(params[:id])
 
@@ -74,11 +77,13 @@ module Api
 
         evidence_loc = project.portfolio_path
 
-        if evidence_loc.nil? || File.exist?(evidence_loc) == false
-          evidence_loc = Rails.root.join('public', 'resources', 'FileNotFound.pdf')
-          header['Content-Disposition'] = 'attachment; filename=FileNotFound.pdf'
-        else
-          header['Content-Disposition'] = 'attachment; filename=portfolio.pdf'
+        if params[:as_attachment]
+          if evidence_loc.nil? || File.exist?(evidence_loc) == false
+            evidence_loc = Rails.root.join('public', 'resources', 'FileNotFound.pdf')
+            header['Content-Disposition'] = 'attachment; filename=FileNotFound.pdf'
+          else
+            header['Content-Disposition'] = 'attachment; filename=portfolio.pdf'
+          end
         end
 
         # Set download headers...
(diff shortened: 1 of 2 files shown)


app/api/submission/portfolio_api.rb
```diff
@@ -74,11 +77,13 @@ class PortfolioApi < Grape::API
 
         evidence_loc = project.portfolio_path
 
-        if evidence_loc.nil? || File.exist?(evidence_loc) == false
-          evidence_loc = Rails.root.join('public', 'resources', 'FileNotFound.pdf')
-          header['Content-Disposition'] = 'attachment; filename=FileNotFound.pdf'
-        else
-          header['Content-Disposition'] = 'attachment; filename=portfolio.pdf'
+        if params[:as_attachment]
+          if evidence_loc.nil? || File.exist?(evidence_loc) == false
```
macite: We need to pull this test out of the attachment test. This checks if the file does not exist and switches it to use the FileNotFound.pdf instead. This way the end point always returns some PDF.

app/api/submission/portfolio_api.rb
```diff
@@ -74,11 +77,13 @@ class PortfolioApi < Grape::API
...
+            evidence_loc = Rails.root.join('public', 'resources', 'FileNotFound.pdf')
```
macite: Keep this outside the attachment test

app/api/submission/portfolio_api.rb
```diff
@@ -74,11 +77,13 @@ class PortfolioApi < Grape::API
...
+            header['Content-Disposition'] = 'attachment; filename=FileNotFound.pdf'
```
macite: I suggest we add a `filename` variable - set outside the attachment test.
Then within the attachment test if we can set the content disposition to use the filename from the variable. Then there is no need for the inner if here...

app/api/submission/portfolio_evidence_api.rb
```diff
@@ -68,18 +71,20 @@ class PortfolioEvidenceApi < Grape::API
...
-          evidence_loc = Rails.root.join('public', 'resources', 'AwaitingProcessing.pdf')
-          header['Content-Disposition'] = 'attachment; filename=AwaitingProcessing.pdf'
-        elsif evidence_loc.nil?
-          evidence_loc = Rails.root.join('public', 'resources', 'FileNotFound.pdf')
-          header['Content-Disposition'] = 'attachment; filename=FileNotFound.pdf'
-        else
-          header['Content-Disposition'] = "attachment; filename=#{task.task_definition.abbreviation}.pdf"
+        if params[:as_attachment]
+          if task.processing_pdf?
+            evidence_loc = Rails.root.join('public', 'resources', 'AwaitingProcessing.pdf')
```
macite: Same as above - need to set `evidence_loc` outside of the attachment if test.

PR #78, opened 2018-05-20
NEW: Teaching Periods

AkshatBajaj:
Standardise the process of unit creation

### DIFF
--- a/app/models/unit.rb
+++ b/app/models/unit.rb
@@ -110,7 +110,7 @@ class Unit < ActiveRecord::Base
     if teaching_period_id.present?
       tp = TeachingPeriod.find(teaching_period_id)
       write_attribute(:start_date, tp.start_date)
-      write_attribute(:end_date, tp.end_date)      
+      write_attribute(:end_date, tp.end_date)
     end
     super(teaching_period_id)
   end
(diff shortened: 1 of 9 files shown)


app/models/unit.rb
```diff
@@ -91,6 +91,9 @@ def role_for(user)
   has_many :convenors, -> { joins(:role).where('roles.name = :role', role: 'Convenor') }, class_name: 'UnitRole'
   has_many :staff, ->     { joins(:role).where('roles.name = :role_convenor or roles.name = :role_tutor', role_convenor: 'Convenor', role_tutor: 'Tutor') }, class_name: 'UnitRole'
 
+  # Unit has a teaching period
+  belongs_to :teaching_period
+
```
macite: I think we should also update the model to define start_date and start_date= as well as end_date and end_date=

In these you can access self's attributes for start date/end date or the ones from the teaching period. In the "=" variations throw an exception if there is a teaching period.

Then update the API end point - no need to assign to start and end date as these should be read through the relationship. Then if someone changes the start and end date of the trimester all units will change as well.

macite: We do this kind of thing with the json attributes in the TaskDefinition if you want an example.

PR #81, opened 2018-07-17
FIX: Check if path exists before removing in zip submissions

jakerenzella:
This PR fixes a bug which occurs when a zip of all submission files is generated for empty sets


jakerenzella:
@macite Resolved 👍

PR #82, opened 2018-07-17
ENHANCE: Replace terminator with Ruby native calls

jakerenzella:
This PR removes the use of Terminator with simpler Ruby calls.


macite:
Manually merged, with additions to remove gem from gem file

PR #85, opened 2018-07-26
Feature/add teaching period

AkshatBajaj:
This pull request will standardise the process of unit creation through teaching periods.


app/models/teaching_period.rb
```diff
@@ -0,0 +1,17 @@
+class TeachingPeriod < ActiveRecord::Base
+  has_many :units
+
+  validates :period, length: { minimum: 1, maximum: 20, allow_blank: false }, uniqueness: true
+  validates :year, length: { is: 4, allow_blank: false }, presence: true, numericality: { only_integer: true },
+    inclusion: { in: 2000..2099, message: "%{value} is not a valid year" }
```
macite: Lets give us a 1000 year window :) make it 2999

app/models/teaching_period.rb
```diff
@@ -0,0 +1,17 @@
+class TeachingPeriod < ActiveRecord::Base
```
macite: Dont allow teaching periods to be destroyed if they have units

app/api/teaching_periods.rb
```diff
@@ -0,0 +1,124 @@
...
+      teaching_period_id = params[:teaching_period_id]
+      TeachingPeriod.find(teaching_period_id).destroy
+    end
+
+    # Returns the current units with teaching periods
+    # Return:
+    # Unit Name, unit ID, teaching period name, year
+    #
+    desc 'Return all the units teaching period information'
+    get '/teaching_periods/id/units' do
```
macite: What is this for?

I can imagine getting all of the unit for a given teaching period... but not all with a teaching period set at all...

app/models/unit.rb
```diff
@@ -92,13 +92,49 @@ def role_for(user)
   has_many :convenors, -> { joins(:role).where('roles.name = :role', role: 'Convenor') }, class_name: 'UnitRole'
   has_many :staff, ->     { joins(:role).where('roles.name = :role_convenor or roles.name = :role_tutor', role_convenor: 'Convenor', role_tutor: 'Tutor') }, class_name: 'UnitRole'
 
+  # Unit has a teaching period
+  belongs_to :teaching_period
+  validates :start_date, presence: true
+  validates :end_date, presence: true
+
+  validate :validate_end_date_after_start_date
+  validate :ensure_teaching_period_dates_match, if: :has_teaching_period?
```
macite: Can we not override start_date and end_date getters so that they access the teaching period value if it is there?

app/models/unit.rb
```diff
@@ -92,13 +92,49 @@ def role_for(user)
...
+      write_attribute(:end_date, tp.end_date)
+    end
+    super(teaching_period_id)
+  end
+
+  def has_teaching_period?
+    self.teaching_period.present?
+  end
+
+  def ensure_teaching_period_dates_match
```
macite: If this fails... what can the user do?

app/api/teaching_periods.rb
```diff
@@ -0,0 +1,124 @@
...
+      teaching_period.update!(teaching_period_parameters)
+      teaching_period
+    end
+
+    desc 'Get all the Teaching Periods'
+    get '/teaching_periods' do
+      unless authorise? current_user, User, :get_teaching_periods
+        error!({ error: 'Not authorised to get teaching periods' }, 403)
+      end
+      teaching_periods = TeachingPeriod.all
```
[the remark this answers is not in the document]
AkshatBajaj: Yes, at the moment students cannot see the teaching period details. It is a good idea, I can add another association to the Teaching Period.

app/api/teaching_periods.rb
AkshatBajaj: This is for the e-solution project, it is basically returning the units that are associated with a teaching period.

app/models/unit.rb
AkshatBajaj: I added that, but you asked me to remove it because it was redundant. If there is a teaching period, then I am manually writing the attribute value, so it is getting the teaching period value if it is there.
If you want, we can still add it.

app/models/unit.rb
AkshatBajaj: If this fails, the API will throw an error.
Practically, this will never happen, because whenever there is a teaching period, I am manually overriding the value of start and end date to make it same.

app/api/teaching_periods.rb
macite: Ok - then it needs to be un-authenticated. As this will be accessed without a user login.

macite: Is that the data they need? Can you check that?

app/models/unit.rb
macite: Ok - lets leave this as is now.

app/models/unit.rb
macite: This should probably just be in a unit test then...

macite: May be better to override the setters for the dates and raise errors there if they are set when there is a teaching period... though this complicates the setting of these when we do set the teaching period :)

app/api/teaching_periods.rb
AkshatBajaj: Jake asked me to add this, I will confirm from him when he comes back.

app/models/teaching_period.rb
AkshatBajaj: Done :+1:

app/models/teaching_period.rb
```diff
@@ -0,0 +1,17 @@
...
+  has_many :units
+
+  validates :period, length: { minimum: 1, maximum: 20, allow_blank: false }, uniqueness: true
```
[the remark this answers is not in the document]
AkshatBajaj: Done :)
validate

app/api/teaching_periods.rb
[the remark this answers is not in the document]
AkshatBajaj: Done :)

PR #93, opened 2018-08-17
Fix #92, add byebug gem

### LINKED ISSUE #92: BUG: User email regex accepting invalid emails
The email address regex for users API doesn’t seem to be working properly. I’m able to create accounts with these emails: `qwertyuiop@qwe qwertyuiop@.com qwertyuiop@blah..com`. I haven't tried more combinations at the moment.
The bug can be reproduced using Swagger or the follow minitest:
```
  def test_post_create_user_invalid_email
    pre_count = User.all.length
    user = {
        first_name: 'Akash',
        last_name: 'Agarwal',
        email: 'blah@blah.com',
        username: 'akash',
        nickname: 'akash',
        system_role: 'Admin'
    }

    invalid_emails = %w(qwertyuiop qwertyuiop@qwe qwertyuiop@.com qwertyuiop@blah..com)

    invalid_emails.each do |email|
      # Assign invalid email
      user[:email] = email

      data_to_post = {
          user: user,
          auth_token: auth_token
      }

      post_json '/api/users', data_to_post
      # Successful assertion of same length again means no record was created
      assert_equal pre_count, User.all.length
      assert_equal 500, last_response.status
    end
  end
```

akashagarwal7:


app/api/users.rb
```diff
@@ -115,6 +115,10 @@ class Users < Grape::API
           end
           action = new_role.id > old_role.id ? :promote_user : :demote_user
 
+          unless user_parameters[:email] =~ /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
```
macite: Can we move this into the User model?

You can add a custom validation there which will cover any modification to the field - and raise exceptions on error.

akashagarwal7:
@macite I didn't remove the following email piece of code from `import_from_csv(current_user, file)` method under `models/user.rb` as per our discussion on Slack since the errors are sent back to the user.

```ruby
unless email =~ /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
  errors << { row: row, message: "Invalid email address (#{email})" }
  next
end
```

PR #94, opened 2018-08-27
Travis CI integration

akashagarwal7:
Works for both staging and testing environment


macite:
Can we fix the use of random in the test that is failing? No need to select random elements in any case.

akashagarwal7:
@macite Done!

PR #95, opened 2018-08-30
FIX: Ensure submission exists before cleanup

jakerenzella:
This PR will ensure that the task def clean up when there are group submissions will succeed


macite:
I fixed this with some additional changes along these lines

PR #125, opened 2018-11-15
Remove an unavailable gem minitest-osx

ohtake:
'minitest-osx' gem has been removed from RubyGems.

https://rubygems.org/gems/minitest-osx
> This gem previously existed, but has been removed by its owner.

Console output:

```
Step 16/19 : RUN bundle install --without production replica
 ---> Running in 651952df1e09
Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
installing your bundle as root will break this application for all non-root
users on this machine.
Fetching gem metadata from https://rubygems.org/.............
Could not find minitest-osx-0.1.0 in any of the sources
ERROR: Service 'api' failed to build: The command '/bin/sh -c bundle install --without production replica' returned a non-zero code: 7
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,6 @@ group :development, :test do
   gem 'factory_girl_rails'
   gem 'minitest-around'
   gem 'minitest-hyper'
-  gem 'minitest-osx'
   gem 'minitest-rails'
   gem 'byebug'
   gem 'simplecov', require: false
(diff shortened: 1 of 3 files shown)


macite:
Can we keep that but make it platform dependent?

There is an example of this here:
https://developers.supportbee.com/blog/using-platform-dependent-gems-rb-inotify-and-rb-fsevent/

ohtake:
The issue is not related to platforms.
The gem has been removed.
It is impossible to build the app on macOS too.

macite:
That is a new... thanks for the fix!

PR #129, opened 2019-02-05
Raise error when no break is added

AkshatBajaj:


macite:
Fix the broken test before we merge

PR #132, opened 2019-02-14
Issue#86: Enrolling a student indicates student id - but not used in …

Aalgham:
…search

update student query to include student number


macite:
Sorry, there is no student_num... and this should be fixed in the front end not the back end!

PR #145, opened 2019-05-25
Add student viewed count

ahujaba:
students could see how many time they have visited the assignment task sheet.


macite:
We can count accesses within the API already - this shouldn't need a new endpoint and the front end to do the calculations.

PR #180, opened 2019-08-29
Fix issue with rolling over units with group tasks

macite:
This PR fixes an issue where group tasks are incorrectly rolled over between teaching periods.

- Update FactoryGirl to correctly create units with groups
- Remove the hyper gem as the reports are not used
- Add a unit test to check that group tasks are correctly rolled over
- Move factory classes into new test structure


macite:
Just realised this also includes a change to the TaskCompletion csv to include the grade and rationale. This change means that we have one export that includes all of the details from the unit. Currently unit chairs need to combine the grade from one with the task completion in another. I am also creating an Excel workbook that will make it easier to explore these values.

PR #200, opened 2019-10-08
Update  TEST: added test case for auth_test.rb

saamgudipudi:
# Added get test commands from 151 to 172.


saamgudipudi:
Modified and improved code coverage of (authentication) auth_test.rb test file

PR #234, opened 2019-12-09
CONFIG: Switch from FactoryGirl to FactoryBot

llyyer:
# Description

CONFIG: Switch from FactoryGirl to FactoryBot

please Combine these 2 commits:Temporary record keeping and CONFIG: SWITCH from FactoryGirl to FactoryBot


macite:
Can you make this PR into the **macite** fork to the **config/new-rails** branch?

Here: https://github.com/macite/doubtfire-api/tree/config/new-rails

PR #236, opened 2019-12-10
CONFIG: Switch from FactoryGirl to FactoryBot

llyyer:
# Description
CONFIG: Switch from FactoryGirl to FactoryBot
please Combine these 2 commits: "Temporary record keeping" and "CONFIG: SWITCH from FactoryGirl to FactoryBot"
The commit "Add test file for controllers" is not necessary, this is the last trimester work.


macite:
It looks like this changes some aspects of the schema that should not change, and adds in the old controller test.

Talk to the tech advisors to get some git related help to separate these out before recreating this PR.

PR #237, opened 2019-12-11
CONFIG: Switch from FactoryGirl to FactoryBot in config/new-rails branch

llyyer:
# Description

Switch from FactoryGirl to FactoryBot

Fixes # (issue)

"FIX: Fix serializers regarding .object", Linh and i fix this error together.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)


test/api/groups_test.rb
```diff
@@ -18,7 +18,8 @@ def app
 
   def test_group_submission_with_extensions
     unit = Unit.first
-
+# print("display:",unit)
```
macite: Make sure to remove these changes - looks like an accidental commit.

db/schema.rb
```diff
@@ -228,15 +228,15 @@
...
-    t.string "type"
-    t.datetime "time_discussion_started"
-    t.datetime "time_discussion_completed"
-    t.integer "number_of_prompts"
-    t.datetime "date_extension_assessed"
-    t.boolean "extension_granted"
-    t.integer "assessor_id"
-    t.integer "task_status_id"
-    t.integer "extension_weeks"
+    t.string "type"
```
macite: Revert these changes - it is just white space so no actual change to this file.

macite: Also check why this is failing in CI...

llyyer: Hello Andrew, due to i delete something at first, then i make the missing things back to make it as same as the original one.if i revert it, if the file will missing something as before?

llyyer: Or i can drop the worng commit, fix it and repull it

macite:
CI is still failing... looks like you need to tell travis to use bundler 2 - https://docs.travis-ci.com/user/languages/ruby/#bundler

llyyer:
> CI仍在失败...看起来您需要告诉travis使用捆绑软件2- [https:](https://docs.travis-ci.com/user/languages/ruby/#bundler) //docs.travis-ci.com/user/languages/ruby/#bundler[](https://docs.travis-ci.com/user/languages/ruby/#bundler)

now, the error seems in Mysql2::Error: Cannot add foreign key constraint, does this mean i should do some works on Mysql2?

macite:
Great! This is a new bug.

Read here: https://ridingtheclutch.com/post/160099545985/rails-51-new-default-bigint-primary-key-sort-of

Then work out the options for us, and come back with a proposal. int is probably enough for now... but maybe we use this to go to bigint for all ids.

macite:
We do use mysql in production at the moment, so we need to get this working.

macite:
@llyyer can you merge in the current changes made to the development branch? These will need to be updated to use FactoryBot now as well. It would be good if we could update the main development branch to be using FactoryBot as well now. You should be able to move across individual commits to make this possible. Can you look into this?

llyyer:
> @llyyer can you merge in the current changes made to the development branch? These will need to be updated to use FactoryBot now as well. It would be good if we could update the main development branch to be using FactoryBot as well now. You should be able to move across individual commits to make this possible. Can you look into this?

do you mean that i pull a new request to the development branch? To my individual development branch or in this development branch?

llyyer:
@macite Hello Andrew, maybe i solve the bigint error, i am not sure if it is.
But there are a new problem, "NoMethodError: undefined method `tutorial_id=' for #<Project:0x0000000009824458>"

i can not find where is this code, can you give me some tips?

PR #250, opened 2020-01-13
TEST: Add Tests to Breaks_API

AreejAsiri:
# Description

TEST: Add Tests for breaks_api by Areej Ali A Asiri

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings


test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+include TestHelpers::JsonHelper
+
+def app
+    Rails.application
+end
+
+# GET TEST 
+# Get All Breaks
+def test_get_all_breaks
+    user = User.admins.first
```
macite: Switch to use a Factory to create any data needed for the test.

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    get with_auth_token('/api/breaks',user)
```
macite: Can you check where this is used in the web project? I have a feeling this endpoint is not needed, in which case we should delete the endpoint rather than creating tests for it.

Look for a resource plus entity for the break url.

test/api/breaks_api_test.rb
macite: So you will need to create a user and a list of breaks...

test/api/breaks_api_test.rb
macite: Do you need a user to access this endpoint? Is it authenticated?

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    expected_data = Break.all
```
macite: Here you can use the list of breaks you create.

test/api/breaks_api_test.rb
SamAnast: Hi Andrew,

I've taken a look at the web project and i can't find any use for this endpoint.
According to [this](https://github.com/doubtfire-lms/doubtfire-web/blob/development/src/app/api/models/break.coffee) we access breaks via teaching periods.

macite: Great - you can remove the test and delete the endpoint code.

linhhtr: @areejasiri please click resolve on this one

linhhtr:
@AreejAsiri when u pushed GET test, u accidentally delete the POST test commit. Please check!

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,61 @@
+require 'test_helper'
+
+class BreaksApiTest < ActiveSupport::TestCase
+include Rack::Test::Methods
```
macite: Indentation...

app/api/breaks_api.rb
```diff
@@ -58,24 +58,6 @@ class BreaksApi < Grape::API
...
-    desc 'Get all the breaks'
-    get '/breaks' do
-      unless authorise? current_user, User, :get_teaching_periods
-        error!({ error: 'Not authorised to get breaks' }, 403)
-      end
-      result = Break.all
-      result
-    end
-
     desc 'Delete a break'
```
macite: This end point isn't used either - so it can go too.

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  assert_equal data_to_post[:start_date].to_date, breaks.start_date.to_date
+  assert_equal data_to_post[:number_of_weeks], breaks.number_of_weeks
+
+   # check if one more break is created
+  assert_equal Break.count, number_of_break + 1
+end 
+
+# GET TEST
+# Get breaks in a teaching period
+def test_unit_main_convenor_get_breaks_in_a_Teaching_Period
```
macite: case!

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    
+    # check if one more break is created
+    assert_equal Break.count, number_of_break + 1
+  end 
+  
+  # GET TEST
+  # Get breaks in a teaching period
+  def test_get_breaks
+    # Create a dummy unit and set teaching period for the unit
+    unit = FactoryBot.create(:unit, teaching_period: FactoryBot.create(:teaching_period))
```
macite: Why create a unit you do not need or use?
Just create the teaching period.

macite: Also - the teaching period factory does not add a break. You will need to update this.

macite: You will also want to tell the factory how many breaks to add (make 0 default)
Then you can test that the right number of breaks exist and they have the right details.

test/api/breaks_api_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    teaching_period = unit.teaching_period
+    
+    # Perform the GET with unit main convenor user
+    get with_auth_token("/api/teaching_periods/#{teaching_period.id}/breaks", unit.main_convenor_user)
+    expected_data = teaching_period.breaks
```
macite: This will currently be empty -- so not really testing anything. You need to think about this,

PR #254, opened 2020-01-19
QUALITY: Adjust tests to use factorygirl data instead of existing data

linhhtr:
# Description

Adjust some existing tests that are using system data to use data generated by FactoryGirl instead to avoid database re-populated; assign correct users to API calls

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/test/api/activity_types_api_test.rb
+++ b/test/api/activity_types_api_test.rb
@@ -51,18 +51,30 @@ class ActivityTypesApiTest < ActiveSupport::TestCase
     assert_equal data_to_post[:activity_type]['abbreviation'], activity_type.abbreviation
   end
 
+  # PUT tests
+  # Replace an activity type
   def test_put_activity_types
+    # a dummy activity type
+    activity_type = FactoryGirl.create(:activity_type)
+    
     data_to_put = {
       activity_type: FactoryGirl.build(:activity_type),
       auth_token: auth_token
     }
 
-    # Update activity_type with id = 1
-    put_json '/api/activity_types/1', data_to_put
+    # Update activity_type with data_to_put
+    put_json "/api/activity_types/#{activity_type.id}", data_to_put
+    
+    #check if the request get through
     assert_equal 200, last_response.status
 
+    # check if the details posted match as expected
     response_keys = %w(name abbreviation)
-    first_activity_type = ActivityType.first
-    assert_json_matches_model(last_response_body, first_activity_type, response_keys)
+    activity_type_updated = activity_type.reload
+    assert_json_matches_model(last_response_body, activity_type_updated, response_keys)
+
+    # check if the details in the replaced teaching period match as data set to replace 
+    assert_equal data_to_put[:activity_type]['name'], activity_type_updated.name
+    assert_equal data_to_put[:activity_type]['abbreviation'], activity_type_updated.abbreviation
   end
 end
--- a/test/api/activity_types_api_test.rb
+++ b/test/api/activity_types_api_test.rb
@@ -23,17 +23,32 @@ class ActivityTypesApiTest < ActiveSupport::TestCase
     end
   end
 
+  # POST tests
+  # Create a new activity type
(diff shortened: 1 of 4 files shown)


test/api/activity_types_api_test.rb
```diff
@@ -23,31 +23,58 @@ def test_get_all_activity_types
     end
   end
 
+  # POST tests
+  # Create a new activity type
   def test_post_activity_types
+    # the number of teaching period before post
+    no_activity_type = ActivityType.count
```
macite: This count isn't used.

test/api/activity_types_api_test.rb
```diff
@@ -23,31 +23,58 @@ def test_get_all_activity_types
...
+
+    # the data that we want to post/create
     data_to_post = {
       activity_type: FactoryGirl.build(:activity_type),
```
macite: Make sure to pull in new upstream changes.

test/api/tutorials_test.rb
```diff
@@ -54,26 +43,15 @@ def test_tutorials_post
...
-      capacity: 10,
-      abbreviation: 'LA011',
-      meeting_location: 'LAB34',
-      meeting_day: 'Tuesday',
-      meeting_time: '18:00'
-    }
-
     data_to_post = {
-      tutorial: tutorial,
+      tutorial: FactoryGirl.build(:tutorial),
```
macite: You will want to pass in the related unit and other details otherwise this will create new units etc.

macite: There are multiple places this applies

test/api/tutorials_test.rb
```diff
@@ -112,11 +79,13 @@ def test_tutorial_post_empty_auth_token
   #4: Testing for failure due to string as Unit ID
   # POST /api/tutorials
   def test_tutorial_post_string_unit_id
-
+    tutor = FactoryGirl.create(:user, :tutor)
+    campus = FactoryGirl.create(:campus)
+    
     tutorial = {
```
macite: This may be a better approach then using FactoryBot.build

test/api/activity_types_api_test.rb
```diff
@@ -23,31 +23,155 @@ def test_get_all_activity_types
...
+    assert_equal no_activity_type +1, ActivityType.count
+  end
+
+  # 2: Convenor cannot create a new activity type
+  def test_convenor_cannot_post_activity_type
+    # Convenor user
+    convenor = FactoryBot.create(:user, :convenor)
+    
+    # the number of teaching period before post
+    no_activity_type = ActivityType.count
```
macite: This is not used in this test

test/api/tutorials_test.rb
```diff
@@ -30,13 +31,48 @@ def assert_tutorial_model_response(response, expected)
...
+      meeting_time: '18:00'
+    }
+
+    data_to_post = {
+      tutorial: tutorial,
+    }
     number_of_tutorials = Tutorial.all.length
 
+    # perform the post with the unit first tutor auth token
+    post_json '/api/tutorials', with_auth_token(data_to_post, tutor)
```
macite: Tutors should not be able to post - my guess is that this tutor is also a convenor.

macite: You need to create a dedicated tutor user and add them to the unit

test/api/tutorials_test.rb
```diff
@@ -102,7 +180,7 @@ def test_tutorial_post_empty_auth_token
       tutorial: tutorial,
       auth_token: ''
     }
-    # perform the post
+    # perform the post with the unit main convenor auth token
     post_json '/api/tutorials', data_to_post
```
macite: Need to set auth token explicitly.

macite:
Well done. This is looking good. It would be good to progress this further now so that we can remove the need to run the populate rake task in testing.

PR #255, opened 2020-01-20
New: Comment replies

jakerenzella:
# Description

Adds backend support for reply_ids

Fixes # (issue)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

All tests pass

# Checklist:

### DIFF
--- a/app/api/task_comments_api.rb
+++ b/app/api/task_comments_api.rb
@@ -13,6 +13,7 @@ module Api
     params do
       optional :comment, type: String, desc: 'The comment text to add to the task'
       optional :attachment, type: Rack::Multipart::UploadedFile, desc: 'Image, sound, PDF or video comment file'
+      optional :reply_to_id, type: Integer, desc: 'The comment to which this comment is replying'
     end
     post '/projects/:project_id/task_def_id/:task_definition_id/comments' do
       project = Project.find(params[:project_id])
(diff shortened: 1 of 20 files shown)


jakerenzella:
@macite WIll add tests, the ID is used to create a hard link to the original comment on the frontend. If the original comment becomes invalid (such that it is deleted), it is handled in the frontend. Furthermore, we can link to things like audio comments this way.

macite:
Why is it a string in the backend then - if it is the id?

jakerenzella:
@macite It is an integer now, that was a mistake -- and added a simple test for the reply_id in the comment_test suite, let me know if we need more here

macite:
Can we add tests for:

- Reply to non-existant id
- Delete of replied to post (what happens to the reply?)
- Reply to id from another task/unit/user etc

Remember to tag me for review - as there are many open PRs and I'm only checking where I am marked for review.

jakerenzella:
@macite Please review changes 👍

app/api/task_comments_api.rb
```diff
@@ -13,6 +14,7 @@ class TaskCommentsApi < Grape::API
     params do
       optional :comment, type: String, desc: 'The comment text to add to the task'
       optional :attachment, type: Rack::Multipart::UploadedFile, desc: 'Image, sound, PDF or video comment file'
+      optional :reply_to, type: String, desc: 'The comment to which this comment is replying'
```
macite: Can use integer type here to ensure it is a number.

test/api/comments/comment_test.rb
```diff
@@ -29,20 +29,80 @@ def test_student_post_comment
     assert_equal pre_count + 1, TaskComment.count, "one comment added"
 
     expected_response = {
-        "comment" => "Hello World",
-        "has_attachment" => false,
-        "type" => "text",
-        "is_new" => false,
-        author: {"id" => user.id},
-        recipient: {"id" => tutor.id}
+      "comment" => "Hello World",
```
macite: Could update string types here to '

test/api/comments/comment_test.rb
```diff
@@ -29,20 +29,80 @@ def test_student_post_comment
...
 
     # check each is the same
-    assert_json_matches_model last_response_body, expected_response, ["comment", "has_attachment", "type", "is_new"]
+    assert_json_matches_model last_response_body, expected_response, %w(comment has_attachment type is_new)
     assert_json_matches_model last_response_body["author"], expected_response[:author], ["id"]
     assert_json_matches_model last_response_body["recipient"], expected_response[:recipient], ["id"]
   end
 
+  def test_replying_to_comments
+    project = Project.first
```
macite: Don't use existing data - use FactoryGirl to create.

test/api/comments/comment_test.rb
```diff
@@ -29,20 +29,80 @@ def test_student_post_comment
...
+    comment_data = { comment: "Responding again!", reply_to: TaskComment.last.id }
+    post_json with_auth_token("/api/projects/#{project.id}/task_def_id/#{task_definition.id}/comments", tutor), comment_data
+    assert_equal 201, last_response.status
+
+    # check each is the same
+    assert_json_matches_model last_response_body, expected_response, %w(comment type is_new reply_to)
+  end
+
+  def test_student_post_reply_to_invalid_comment
+    project = Project.first
```
macite: Same here - FactoryGirl.create

test/api/comments/comment_test.rb
```diff
@@ -29,20 +29,80 @@ def test_student_post_comment
...
+    unit = project.unit
+    task_definition = unit.task_definitions.first
+
+    comment_data = { comment: "Responding!", reply_to: -1 }
+
+    post_json with_auth_token("/api/projects/#{project.id}/task_def_id/#{task_definition.id}/comments", user), comment_data
+
+    assert_equal 404, last_response.status
+  end
+
```
macite: Can we also test:

- reply to comment not in unit - so valid comment id but should not be accessible
- reply to other students comment - so valid comment id but should not be accessible
- reply to other student comment in a group task (where valid [same team] and not valid [other team])
- also tutor reply to a different student comment in other student task - so valid comment, with access, but in unrelated student's task
- also what about reply to a comment in a different task for the same student - I assume this should not be allowed.

macite: These should all be enforced via active model validations.

macite:
Remember to re-request review also - as there are lots of open PRs I am not only checking those awaiting my review.

jakerenzella:
@macite I think this is 95% there... All cases you mentioned are now tested however having troubles with a student from the same group failing to be able to comment on a task, this is failing regardless of the reply to I am sure it is something I am doing wrong with the setup of the test. I have left the test commented for feedback. The rest should be good. reply_to_id is now an actual reference to a TaskComment, validation is added to essentially be if you can get a comment, you can reply to it.

jakerenzella:
@macite I know this isn't a priority but could you take a look at this? I believe the CI failing is just because it requires a migration.

Angular 9 is ready for upgrade and I'd like to get as many PRs merged before working on that migration 👍

PR #285, opened 2020-06-01
NEW: Add submitted grade to portfolio submission

C-Blenco:
# Description

This feature adds a new grade type for the student, the "submitted grade". Now when a student submits a portfolio, the grade select buttons will modify their "submitted grade", rather than their "target grade". When marking a portfolio, an assessor can now see their target grade, and the grade they submitted the portfolio as (submitted grade).
Please refer to doubtfire-web PR for a screenshot of the implementation: [NEW: Add submitted grade to portfolio submission](https://github.com/doubtfire-lms/doubtfire-web/pull/280)

This PR changes the following:

- A new submitted_grade column in the projects table
- Return submitted grade value when project details are requested

## Type of change

# How Has This Been Tested?

A test was written to ensure that the submitted grade value can't be modified after the student submits the portfolio. The test also tests a successful change in the submitted grade value before submission.

- [x] Test that submitted grade can't be modified after portfolio submission

# Checklist:


app/models/project.rb
```diff
@@ -311,6 +311,10 @@ def target_grade=(value)
     self[:target_grade] = value
   end
 
+  def submitted_grade=(value)
```
macite, line 314: Do we really need this? Generally you should be able to do this without needing to actually write this code as ActiveRecord should handle it.

PR #287, opened 2020-08-18
NEW: Allow exporting task dates in iCalendar format

RavinduL:
# Description

Adds functionality that serves key task dates of units that a user is enrolled in, as an iCalendar, importable into Outlook/Google calendars.

The corresponding frontend PR is: https://github.com/doubtfire-lms/doubtfire-web/pull/293

## Type of change

# How Has This Been Tested?

Unit tests ensure that webcals,

- Include only task definitions of current units
- Include only task definitions that are targeted
- Include only task definitions of units that aren't excluded
- Include events with target dates of all task definitions
- Include events for start & end dates if include_start_dates
- Include events with extended date if available

...and that, of the API,

- Setting enabled creates or destroys webcal
- Enables with sensible defaults
- should_change_guid parameter changes the Webcal guid
- Ical endpoint is public and serves webcal with corect content type
- Reminder must be specified with both time & unit

image

# Checklist:

### DIFF
--- a/app/api/webcal_api.rb
+++ b/app/api/webcal_api.rb
@@ -9,13 +9,23 @@ module Api
     # Declare content types
     content_type :txt, 'text/calendar'
 
+    helpers do
+      #
+      # Wraps the specified value (expected to be either `nil` or a `Webcal`) in a hash `{ enabled: true | false }` used
+      # to prevent the API returning `null`.
+      #
+      def wrap_webcal(webcal)
+        { enabled: webcal.present? }.merge(webcal.present? ? WebcalSerializer.new(webcal).as_json : {})
+      end
+    end
+
     desc 'Get webcal details of the authenticated user'
     params do
       requires :auth_token, type: String, desc: 'Authentication token'
     end
     get '/webcal' do
       authenticated?
-      current_user.webcal
+      wrap_webcal current_user.webcal
     end
 
     desc 'Update webcal details of the authenticated user'
@@ -56,7 +66,8 @@ module Api
         end
       end
 
-      return if cal.nil? or cal.destroyed?
+      return wrap_webcal(nil) if cal.nil? or cal.destroyed?
+
       webcal_update_params = {}
 
       # Change the GUID if requested.
@@ -105,7 +116,7 @@ module Api
         end
       end
 
-      cal
+      wrap_webcal cal
     end
 
     desc 'Serve webcal with the specified GUID'
--- a/app/api/webcal_api.rb
+++ b/app/api/webcal_api.rb
@@ -81,7 +81,8 @@ module Api
             projects: { user_id: webcal.user_id },
             units: { active: true }
           )
-          .where('? BETWEEN units.start_date AND units.end_date', Time.zone.now)
+          .where('tasks.project_id is null or tasks.project_id = projects.id')   # eager_load(:tasks) only tasks of :projects
+          .where('? BETWEEN units.start_date AND units.end_date', Time.zone.now) # Current units
       )
 
       # Specify refresh interval.
--- a/app/api/webcal_api.rb
+++ b/app/api/webcal_api.rb
@@ -21,9 +21,10 @@ module Api
     desc 'Update webcal details of the authenticated user'
     params do
       requires :webcal, type: Hash do
-        optional :enabled,             type: Boolean, desc: 'Is the webcal enabled?'
-        optional :should_change_guid,  type: Boolean, desc: 'Should the GUID of the webcal be changed?'
-        optional :include_start_dates, type: Boolean, desc: 'Should events for start dates be included?'
+        optional :enabled,             type: Boolean,        desc: 'Is the webcal enabled?'
+        optional :should_change_guid,  type: Boolean,        desc: 'Should the GUID of the webcal be changed?'
+        optional :include_start_dates, type: Boolean,        desc: 'Should events for start dates be included?'
+        optional :unit_exclusions,     type: Array[Integer], desc: 'IDs of units that must be excluded from the webcal'
       end
       requires :auth_token, type: String, desc: 'Authentication token'
     end
@@ -32,7 +33,12 @@ module Api
       webcal_params = params[:webcal]
 
       user = current_user
-      cal = user.webcal
+
+      cal = Webcal
+        .includes(:webcal_unit_exclusions)
+        .where(user_id: user.id)
+        .load
+        .first
 
       # Create or destroy the user's webcal, according to the `enabled` parameter.
       if webcal_params.key?(:enabled)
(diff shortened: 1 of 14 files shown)


RavinduL:
Planned work that's currently pending,

- [ ] Feature demonstration @RavinduL 

- [ ] Unit tests @RavinduL 

- [ ] Fix issue with `webcal#id` being defined implicitly as an integer in [schema.rb](https://github.com/OnTrack-UG-Squad/doubtfire-api/blob/feature/task-webcal/db/schema.rb).

- [ ] Inclusion of extension dates instead of task end dates where applicable @AnkanSubedi

- [ ] Frontend @Yuntannn

db/schema.rb
```diff
@@ -415,6 +415,13 @@
   add_index "users", ["authentication_token"], name: "index_users_on_authentication_token", unique: true, using: :btree
   add_index "users", ["login_id"], name: "index_users_on_login_id", unique: true, using: :btree
 
+  create_table "webcals", force: :cascade do |t|
```
RavinduL, line 418: The migration defines the `webcals` table with a `string` primary key,

```ruby
create_table(:webcals, id: :string) do |t|
  t.boolean    :include_start_dates, null: false, default: false
  t.references :user, foreign_key: true, index: { unique: true }
end
```

...and `db:migrate` migrates it accordingly, but it doesn't get included in `schema.rb` and therefore `db:schema:load` implicitly defines `webcals#id` as an integer. 

https://github.com/rails/rails/issues/16599 is addresses this, but the fix isn't released in Rails 4.2

RavinduL:
The migration defines the `webcals` table with a `string` primary key,

https://github.com/doubtfire-lms/doubtfire-api/blob/7dbcfdd4bd68dfcb3ca04488b525aea2c697adda/db/migrate/20200817020024_create_webcals.rb#L3-L5

...and `db:migrate` migrates it accordingly, but it doesn't get included in `schema.rb`,

https://github.com/doubtfire-lms/doubtfire-api/blob/7dbcfdd4bd68dfcb3ca04488b525aea2c697adda/db/schema.rb#L418-L421

...and therefore `db:schema:load` implicitly defines `webcals#id` as an integer. 

https://github.com/rails/rails/issues/16599 is addresses this, but the fix isn't released in Rails 4.2

app/api/webcal_api.rb
```diff
@@ -0,0 +1,119 @@
...
+    put '/webcal' do
+      authenticated?
+      webcal_params = params[:webcal]
+
+      user = current_user
+      cal = user.webcal
+
+      # Create or destroy the user's webcal, according to the `enabled` parameter.
+      if webcal_params.key?(:enabled)
+        if webcal_params[:enabled] and cal == nil
```
macite, line 39: In ruby use `.nil?` eg `cal.nil?`

app/api/webcal_api.rb
```diff
@@ -0,0 +1,119 @@
...
+          cal = user.create_webcal(id: SecureRandom.uuid)
+        elsif (not webcal_params[:enabled]) and cal != nil
```
macite, line 41: With rails you can use `.present?` rather than `!cal.nil?`

app/api/webcal_api.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      # https://tools.ietf.org/html/rfc7986#section-5.7
+      ical.append_custom_property('REFRESH-INTERVAL', refresh_interval)
+
+      # Retrieve task definitions and tasks of the user's active units.
+      TaskDefinition
+          .eager_load(:tasks)
+          .joins(unit: :projects)
+          .where(
+            projects: { user_id: 7 },
```
macite, line 88: Get the user from the webcal

app/api/webcal_api.rb
```diff
@@ -0,0 +1,119 @@
...
+            units: { active: true }
```
macite, line 89: Are "active" units from previous trimesters excluded here?

app/api/webcal_api.rb
```diff
@@ -0,0 +1,119 @@
...
+    params do
+      requires :id, type: String, desc: 'The ID of the webcal'
+    end
+    get '/webcal/:id' do
+
+      # Retrieve the specified webcal.
+      webcal = Webcal.find(params[:id])
+      ical = Icalendar::Calendar.new
+
+      # Specify refresh interval.
```
macite, line 74: Move this logic into the model object

macite (commented):
Progress looks good.

Needs unit tests

app/api/webcal_api.rb
```diff
@@ -81,7 +81,8 @@ class WebcalApi < Grape::API
             projects: { user_id: webcal.user_id },
             units: { active: true }
           )
-          .where('? BETWEEN units.start_date AND units.end_date', Time.zone.now)
+          .where('tasks.project_id is null or tasks.project_id = projects.id')   # eager_load(:tasks) only tasks of :projects
+          .where('? BETWEEN units.start_date AND units.end_date', Time.zone.now) # Current units
```
RavinduL, line 85: This is equivalent to the `Unit.current` scope.

https://github.com/OnTrack-UG-Squad/doubtfire-api/blob/development/app/models/unit.rb#L148

Is it possible to somehow include `Unit.current` in the join so as to not redefine the scope here?

app/api/webcal_api.rb
```diff
@@ -81,7 +81,8 @@ class WebcalApi < Grape::API
             projects: { user_id: webcal.user_id },
             units: { active: true }
           )
-          .where('? BETWEEN units.start_date AND units.end_date', Time.zone.now)
+          .where('tasks.project_id is null or tasks.project_id = projects.id')   # eager_load(:tasks) only tasks of :projects
```
RavinduL, line 84: This is effectively,

```sql
SELECT *
FROM tasks_definitions td
INNER JOIN projects p ON
    ...
LEFT JOIN tasks t ON
    t.task_definition_id = td.id
WHERE
    (t.project_id is null or t.project_id = projects.id)
```

Is there a Rails equivalent of the following?

```sql
SELECT *
FROM tasks_definitions td
INNER JOIN projects p ON
    ...
LEFT JOIN tasks t ON
    t.task_definition_id = td.id
    AND t.project_id = p.id
```

macite: I've done this elsewhere using `joins('LEFT OUTER JOIN tutorials ON tutorial_enrolments.tutorial_id = tutorials.id AND (tutorials.tutorial_stream_id = task_definitions.tutorial_stream_id OR tutorials.tutorial_stream_id IS NULL)').`

app/api/webcal_api.rb
macite: Not sure this would be beneficial. In the end you are just creating a query here to report data.

app/api/webcal_api.rb
macite: Not sure this would be beneficial. In the end you are just creating a query here to report data.

PR #288, opened 2020-08-22
NEW: Allow students to copy their Draft Learning Summary to portfolio

C-Blenco:
# Description

Adds functionality for unit staff to assign a task definition as a draft learning summary task, and allows a student to copy their uploaded draft learning summary directly to their portfolio.
Corresponding front-end pull request: doubtfire-lms/doubtfire-web#304

## Type of change

# How Has This Been Tested?

Unit tests have been implemented to ensure drafts copy over correctly and validation regarding upload requirements and changes to the selected draft learning summary task.

# Checklist:


app/api/submission/portfolio_api.rb
```diff
@@ -95,6 +95,18 @@ class PortfolioApi < Grape::API
         File.read(evidence_loc)
       end # get
 
+
+      desc "Use draft learning summary upload as portfolio learning summary"
+      post '/submission/project/:id/portfolio/draft_summary' do
```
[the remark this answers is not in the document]
macite: Good idea!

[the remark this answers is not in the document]
macite: Maybe rename the end point to "move_draft_summary"

app/api/submission/portfolio_api.rb
```diff
@@ -95,6 +95,24 @@ class PortfolioApi < Grape::API
...
+        project = Project.find(params[:id])
+
+        unless authorise? current_user, project, :make_submission
+          error!({ error: "Not authorised to alter portfolio for project '#{params[:id]}'" }, 401)
+        end
+
+        result = project.move_draft_to_portfolio
+
+        if result == false
```
macite, line 109: if ! result

app/api/units_api.rb
```diff
@@ -89,13 +90,21 @@ class UnitsApi < Grape::API
...
+                                                            :send_notifications,
+                                                            :draft_task_definition_id
                                                           )
 
       if unit.teaching_period_id.present? && unit_parameters.key?(:start_date)
         unit.teaching_period = nil
       end
 
+      if unit_parameters.key?(:draft_task_definition_id)
+        unless unit.task_definitions.exists?(unit_parameters[:draft_task_definition_id])
```
macite, line 102: should we also check it is a valid task? It should only have a single document upload.

macite: Also need to think about validating changes to tasks - if the draft learning summary is later changed it could break this.

app/models/task_definition.rb
```diff
@@ -15,6 +15,8 @@ class TaskDefinition < ActiveRecord::Base
   belongs_to :group_set
   belongs_to :tutorial_stream
 
+  has_one :draft_task_definition_unit, foreign_key: 'draft_task_definition_id', class_name: 'Unit', dependent: :nullify
```
macite, line 18: This should be the other way around - this may belong to a unit as the draft learning summary.

app/models/unit.rb
```diff
@@ -131,6 +131,8 @@ def role_for(user)
 
   belongs_to :main_convenor, class_name: 'UnitRole'
 
+  belongs_to :draft_task_definition, class_name: 'TaskDefinition'
```
macite, line 134: has one...

app/models/project.rb
```diff
@@ -705,6 +708,39 @@ def move_to_portfolio(file, name, kind)
...
+    return false if task.nil?
+    return false if !task.has_pdf
+
+    # Extract to porfolio temp directory
+    zip_file = task.zip_file_path_for_done_task
+    if zip_file && File.exist?(zip_file)
+      Zip::File.open(zip_file) do |zip|
+        zip.glob("**/*").each do |entry|
+          next if entry.name_is_directory?
+          result[:name] = "DraftLearningSummaryReport.pdf"
```
macite, line 727: Why not move this in place as the learning summary report? No need for a separate "draft". That way it will already be the learning summary report when the portfolio is created. Otherwise you will need to copy the draft to real on creation of the portfolio.

macite (changes requested):
After reviewing I think we have the design a little wrong. Lets make it work more automatically...

1. Student submits draft learning summary
2. PDF is copied into portfolio as learning summary report... unless there is one already there uploaded by the student. (potentially as part of the generation of the PDF)

So we could then drop the end point needed to trigger the copy.
We would need a flag in the project to indicate that the draft was being used
In the UI we could then use the "draft learning summary in use" flag to show a message in the portfolio wizard - indicating the draft is in use.
The flag can also be used when copying the draft - copy draft unless  "draft learning summary in use" is false and there is a PDF already there

Thoughts?

app/models/task_definition.rb
C-Blenco: From my understanding, `has_one` points to a table that contains a foreign key, whereas belongs to will be the class containing the foreign key. Since a Unit has the foreign key of `draft_task_definition_id` wouldn't this be the correct way around?
A `belongs_to` relation also doesn't accept a `:nullify` dependent, which needs to be set on the `TaskDefinition`, since we aim to nullify the FK on the Unit.

I could be understanding it wrong, let me know what you think!

macite: That makes sense :) - just seems backward. Will this cause any issues on destroy of a unit? May be worth checking due to the circular relationship.

C-Blenco:
@macite 
> 1. Student submits draft learning summary
> 2. PDF is copied into portfolio as learning summary report... unless there is one already there uploaded by the student. (potentially as part of the generation of the PDF)

I'm happy to re-write it to function this way. This would require special logic for the draft learning summary task yes? So when generating the PDF, if the `task.task_definition_id == unit.draft_task_definition_id` we would also copy to the Projects portfolio directory?

Another issue might be whether we still want to copy it over to the portfolio if the tutor hasn't marked the task as complete yet (I guess the tutor could say the draft isn't good enough). Do we need to worry about this?

macite:
Yes this would need the task code to be updated. No need to worry about the tutor not signing it off as complete before it can be used. If it gets to the point where we need the learning summary and this is all we have, then that is what we use. The message in the portfolio creation at the front end should strongly encourage the student to upload a new version.

macite:
Is there capacity to look at migrating the portfolio creation Wizard to Angular as part of this?

C-Blenco:
@macite 
> Is there capacity to look at migrating the portfolio creation Wizard to Angular as part of this?
Potentially, I'll have a chat with my squad about this. I'm assuming it would also involve quite a large re-design based on @jakerenzella's updated UI design.

Either way, I'll finish the changes in the back end and discuss the possibility of it with my squad!

RavinduL:
> Is there capacity to look at migrating the portfolio creation Wizard to Angular as part of this?

This'll be fun! @C-Blenco thoughts on doing this as a separate epic (that is, instead of in `draft-learning-summary`)? 🤔

app/models/task_definition.rb
C-Blenco: > just seems backward.

I totally agree, I struggled to wrap my head around it for that very reason, the `belong` and `has` terminology seems as if it would indicate a different relation. I'll do some further testing just incase. 

> Will this cause any issues on destroy of a unit? May be worth checking due to the circular relationship.

I'll definitely test this. When I write some unit tests as well I'll test all of these things

macite (changes requested):
Looks good, just need to expand testing a little.

Test that you cannot make a task with multiple uploads the draft learning summary, and that it must be a document, and that you cannot change it after setting as draft learning summary.

Should be good then.

C-Blenco:
> Test that you cannot make a task with multiple uploads the draft learning summary, and that it must be a document, and that you cannot change it after setting as draft learning summary.

Added these tests and fixed an issue regarding task definition API validation

macite (approved):

PR #295, opened 2020-10-02
FIX: Fix updating unit on units API when draft learning summary id is null or empty

C-Blenco:
# Description

A bug existed where when using `.key?` when detecting if a draft task definition id was passed into the parameters would also equate to `true` when the front-end returned null or an empty string.

## Type of change

# How Has This Been Tested?

The tests previously created in `units_api_tests.rb` have no failures.

# Checklist:

### DIFF
--- a/app/api/units_api.rb
+++ b/app/api/units_api.rb
@@ -102,7 +102,7 @@ module Api
         unit.teaching_period = nil
       end
 
-      if unit_parameters[:draft_task_definition_id]
+      if unit_parameters[:draft_task_definition_id].present?
         # Ensure the task definition belongs to unit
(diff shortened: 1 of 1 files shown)


app/api/units_api.rb
```diff
@@ -102,7 +102,7 @@ class UnitsApi < Grape::API
         unit.teaching_period = nil
       end
 
-      if unit_parameters.key?(:draft_task_definition_id)
+      if unit_parameters[:draft_task_definition_id]
```
macite, line 105: Can you adjust this to use `.present?` to better capture what you are asking here?

macite (changes requested):

app/api/units_api.rb
C-Blenco: No worries, done!

PR #324, opened 2021-08-25
config: update of gem, docker files and ruby  and rails version

Roboromeo1:


Dockerfile
```diff
@@ -2,12 +2,12 @@ FROM ubuntu:18.04
 
 RUN apt-get update && apt-get install -y curl git
 ENV PATH /root/.rbenv/bin:/root/.rbenv/shims:$PATH
-RUN curl -fsSL https://github.com/rbenv/rbenv-installer/raw/master/bin/rbenv-installer | bash
+RUN curl -fsSL https://github.com/rbenv/rbenv-installer/raw/HEAD/bin/rbenv-installer | bash
```
kugamigh, line 5: This seems like an odd place to get the data from.  You might want to change this to main instead of HEAD, as that will be more reliable in terms of knowing which branch you're working with.

Dockerfile
```diff
@@ -34,7 +34,6 @@ ENV PATH /tmp/texlive/bin/x86_64-linux:$PATH
 
 COPY Gemfile Gemfile.lock /doubtfire-api/
 RUN bundle install --without production
-
```
kugamigh, line 37: Is this line a typo/an accidental key press?

macite:
There is a separate branch to update to rails 6 as this breaks much of the system functionality. We need to update the existing branch to incorporate the overseer updates, rather than recreating this here.

macite:
See config/new-rails branch - you will notice it has additional changes that are needed for this update.

Roboromeo1 (commented):
> See config/new-rails branch - you will notice it has additional changes that are needed for this update.

Thank you Prof @macite  i'll check it and work on that

macite:
I've created a new PR - looks like not much change is needed. Though we will need to update the Overseer code. See #325

Dockerfile
kugamigh: Their URL will have changed from:
https://github.com/rbenv/rbenv-installer/raw/master/bin/rbenv-installer

to:
https://github.com/rbenv/rbenv-installer/raw/main/bin/rbenv-installer

PR #330, opened 2021-09-09
config: allow Jupiter notebook file submission

Roboromeo1:


macite:
We need to get ipynb files converted to PDF as part of preprocessing the files.

macite:
in terms of converting tasks to PDF

PR #332, opened 2021-09-14
 Fix : mime types 

Roboromeo1:


app/helpers/file_helper.rb
```diff
@@ -18,8 +18,8 @@ def accept_file(file, name, kind)
       accept = ['image/png', 'image/gif', 'image/bmp', 'image/tiff', 'image/jpeg', 'image/x-ms-bmp']
     when 'code'
       accept = ['text/x-pascal', 'text/x-c', 'text/x-c++', 'text/plain', 'text/', 'application/javascript, text/html',
```
macite, line 20: Can we also fix this here as well?
'application/javascript, text/html' --> missing ' in the middle...

macite:
There is another PR for this as well... which one is the right one?

I'll close this one.

PR #333, opened 2021-09-15
config : Additional Support for system files

Cecilia-Nguyen:


app/models/task.rb
```diff
@@ -962,7 +962,7 @@ def self.pygments_lang(extn)
     elsif ['xml'].include?(extn) then 'xml'
     elsif ['sql'].include?(extn) then 'sql'
     elsif ['vb'].include?(extn) then 'vbnet'
-    elsif ['txt', 'md', 'rmd', 'rpres'].include?(extn) then 'text'
+    elsif ['txt', 'md', 'rmd', 'rpres','hdl','asm','jack','hack','tst','out','cmp','vm','sh','bat','class','dat','jar'].include?(extn) then 'text'
```
macite, line 965: Remove the out file and class and jar... others look good.

macite (changes requested):
Remove the binary file as well...

Then this should be able to be committed.

PR #334, opened 2021-09-16
Update file_helper.rb

kiratkaur13:
added function to receive the python files.


macite:
Needs lots more work...

PR #335, opened 2021-09-18
PDF generation draft initialization

MoeHdaib:
# **PDF Generation Draft PR** 

Initial setup of PDF generation process as we created pdf gen subscriber and a rake task that start the whole process.

Co-authored by @kiratkaur13 

# Current issues and TODO list:

- [ ] Issues with subscriber registration 
- [ ] Code is not yet complete as we are missing some error exceptions
- [ ] Setting up binding keys
- [ ] Test the system by submitting a task through front-end and see if the back-end gets it. 
- [ ] Bug fixes

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -4,6 +4,7 @@ require 'csv'
 require 'yaml'
 require 'grape-active_model_serializers'
 require 'bunny-pub-sub/services_manager'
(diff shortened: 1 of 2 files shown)


macite:
Stale - no progress,

PR #336, opened 2021-09-19
Create apt.txt

kiratkaur13:
this file contains the packages needed for nbconvert


macite:
This would need to be aligned with PDF generation for notebooks.

PR #337, opened 2021-09-19
Create post build

kiratkaur13:
this post build is required to call nbconvert, so that it can convert ipynb files into pdf

- [y ] New feature (non-breaking change which adds functionality)
- [ y] Breaking change (fix or feature that would cause existing functionality to not work as expected)


macite:
Does not look like it is ready to be a contribution yet.

PR #355, opened 2021-12-18
fix: update buf_size in texmf.cnf to fix breaking of pdf generation

Cecilia-Nguyen:
we need to update  buf_size var from 20000 to 400000 in texmf.cnf located  in/tmp/texlive/texmf.cnf inside docker instance.


macite:
This PR does not make a change to the buf_size in the texmf.cnf file.

PR #362, opened 2022-03-10
Perf/fix project and unit load times

macite:
# Description

This PR reinstates the task_stats data for the front end and improves the load time performance of units. This is done by caching the task_stats in the project and updating this on task status change and change of target grade. As a result, this can be loaded from the database on the students query.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -38,6 +38,7 @@ class Project < ApplicationRecord
   validate :tutorial_enrolment_same_campus, if: :will_save_change_to_enrolled?
 
   after_update :check_withdraw_from_groups, if: :saved_change_to_enrolled?
+  after_update :update_task_stats, if: :saved_change_to_target_grade?
 
   #
   # Permissions around project data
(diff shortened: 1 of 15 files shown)


app/models/unit.rb
```diff
@@ -484,7 +465,7 @@ def enrol_student(user, campus)
     Project.create!(
       user_id: user.id,
       unit_id: id,
-      task_stats: '0.0|1.0|0.0|0.0|0.0',
+      task_stats: '0.0|1.0|0.0|0.0|0.0|0.0',
```
tancnle, line 468: **question:** Are we supposed to set this as a JSON string? 🤔 

```rb
{               
  red_pct: 0,   
  grey_pct: 1,  
  orange_pct: 0,
  blue_pct: 0,  
  green_pct: 0, 
  order_scale: 0
}               
```

app/models/task_status.rb
```diff
@@ -62,6 +62,15 @@ def self.time_exceeded
...
   end
 
+  # Return the count (which equals the largest id) - so that other code can loop thought all statuses without database lookup
+  #
+  # Make sure to update this if/when you add another status!
+  #
+  # Keep this hard coded! Saves cache load time.
+  # Important: count must equal the largest id in the database
+  def self.count
+    12
```
tancnle, line 72: **thought:** I think this can be improved. However, it seems that we are hardcoding other index values already (e.g. `TaskStatus.find(12)`), this might be a good compromise for now.

app/models/project.rb
```diff
@@ -568,55 +569,80 @@ def convert_hash_to_pct(hash, total)
...
+          r['count_3'] || 0
+        ]
+      end
+      .first # there is only one row returned...
+
+
+    # Get the assigned tasks (those where task grade <= target grade)
+    result = assigned_tasks
+        .select(
+          *(1..TaskStatus.count).map { |status_id| "SUM(CASE WHEN tasks.task_status_id = #{status_id} THEN 1 ELSE 0 END) AS #{TaskStatus.id_to_key(status_id)}_count" }
```
tancnle, line 627: **suggestion:** This can be extracted to a variable to improve readability. WDYT?

```rb
sum_by_status = (1..TaskStatus.count).map do |status_id|
  "SUM(CASE WHEN tasks.task_status_id = #{status_id} THEN 1 ELSE 0 END) A
S #{TaskStatus.id_to_key(status_id)}_count"
end
```

app/models/project.rb
```diff
@@ -38,6 +38,7 @@ class Project < ApplicationRecord
   validate :tutorial_enrolment_same_campus, if: :will_save_change_to_enrolled?
 
   after_update :check_withdraw_from_groups, if: :saved_change_to_enrolled?
+  after_update :update_task_stats, if: :saved_change_to_target_grade?
```
tancnle, line 41: **thought:** This might increase latency when users try to save the target grades. It could be premature optimisation but we can move this to an async job if it is ever becomes an issue.

tancnle:
Thank you for your hard work on this fix @macite ❤️ I have left some suggestions/questions. Please let me know what you think 🙏🏼

macite:
Thanks for the great review @tancnle 👍 - I will address the issues raised and update ASAP.

app/models/project.rb
macite: For an individual project, the update of task stats takes negligible time - so it should be ok. I'll add a to-do so that we can note this for future performance optimisations.

jakerenzella:
I agree, this is a great review @tancnle  thank you!

app/models/project.rb
```diff
@@ -568,55 +569,80 @@ def convert_hash_to_pct(hash, total)
...
-  #         *TaskStatus.all.map { |s| "SUM(CASE WHEN tasks.task_status_id = #{s.id} THEN 1 ELSE 0 END) AS #{s.status_key}_count" }
-  #       )
-  #       .map do |t| Project.create_task_stats_from(task_count, t, target_grade) end
-  #       .first
-
-  #   if result.nil?
-      {
+  # Recalculate the task stats for the project, and store in the
+  # task_stats field
+  def update_task_stats()
```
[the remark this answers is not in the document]
macite: Important to fix

app/models/project.rb
```diff
@@ -568,55 +569,80 @@ def convert_hash_to_pct(hash, total)
...
+
+    # Get the count of the tasks less than each target grade
+    task_count = unit
+      .task_definitions
+      .select(*(0..3).map { |grade_id| "SUM(CASE WHEN target_grade <= #{grade_id} THEN 1 ELSE 0 END) AS count_#{grade_id}" }) # create columns for each grade
```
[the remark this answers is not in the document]
macite: Agreed - it would be good to have this in here as a constant. It is the max grade index - so probably better as something like grade count - 1

app/models/project.rb
```diff
@@ -568,55 +569,80 @@ def convert_hash_to_pct(hash, total)
...
+      end
+      .first # there is only one row returned...
+
+
+    # Get the assigned tasks (those where task grade <= target grade)
+    result = assigned_tasks
+        .select(
+          *(1..TaskStatus.count).map { |status_id| "SUM(CASE WHEN tasks.task_status_id = #{status_id} THEN 1 ELSE 0 END) AS #{TaskStatus.id_to_key(status_id)}_count" }
+        )
+        .map do |t| Project.create_task_stats_from(task_count, t, target_grade) end
```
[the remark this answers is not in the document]
macite: Yes - lets fix

app/models/project.rb
```diff
@@ -568,55 +569,80 @@ def convert_hash_to_pct(hash, total)
...
+
+    if result.nil?
+      result = {
         red_pct: 0,
         grey_pct: 1,
         orange_pct: 0,
         blue_pct: 0,
         green_pct: 0,
         order_scale: 0
       }
```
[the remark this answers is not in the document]
macite: agreed. Great pickup

db/migrate/20220310092851_update_task_stats.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateTaskStats < ActiveRecord::Migration[7.0]
+  def change
+    # Migrate all projects data - update the task stats
+    Project.all.each do |project|
```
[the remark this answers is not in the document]
macite: Done

app/models/unit.rb
macite: 🤦

macite: Absolutely :)

app/models/task_status.rb
macite: Yes - I was thinking the same. So we will leave this as is... it is probably good to add a TODO comment in there as something to consider later.

app/models/project.rb
macite: Agreed - readability is important

tancnle:
Thank you @macite 🚀

PR #365, opened 2022-03-13
fix:fix the manual delete servers.pid file after restart container

RayGuo-ergou:
# Description

When we restart the doubtfire-api container, we need to delete the ` server.pid ` file in ` /tmp/pids ` manually. So I added a docker entrypoint srcipt to delete that file everytime before the container start if  ` server.pid ` exist. 

## Type of change

- [x] Bug fix

# How Has This Been Tested?

I build the docker image in my local dev environment, then develop the project with that image. To test it, I firstly went to the container bash, make sure the ` server.pid ` file is in the right path then I restart the container several times and restart my PC to test if we still need to delete ` server.pid ` file manually. Also by monitoring the pids folder I can clearly see the file was removed then created when restart the container. 

- [x] restart container
- [x] reboot PC

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -e
+
+if [ -f tmp/pids/server.pid ]; then
+  rm tmp/pids/server.pid
+fi
+
+exec bundle exec "$@"
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,6 +1,0 @@
-#!/bin/bash
-set -e
-
-rm -f tmp/pids/server.pid
-
-exec "$@"
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -3,4 +3,4 @@ set -e
 
 rm -f tmp/pids/server.pid
 
-exec bundle exec "$@"
+exec "$@"
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 set -e
 
-if [ -f tmp/pids/server.pid ]; then
-  rm tmp/pids/server.pid
-fi
+rm -f tmp/pids/server.pid
 
 exec bundle exec "$@"
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,3 +33,7 @@ EXPOSE 3000
 
 ENV RAILS_ENV development
 CMD bundle exec rake db:migrate && bundle exec rails s -b 0.0.0.0
+
+COPY ./docker-entrypoint.sh /docker-entrypoint.sh
+RUN chmod +x /docker-entrypoint.sh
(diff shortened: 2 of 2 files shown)


RayGuo-ergou:
Ref: [https://stackoverflow.com/questions/35022428/rails-server-is-still-running-in-a-new-opened-docker-container](https://stackoverflow.com/questions/35022428/rails-server-is-still-running-in-a-new-opened-docker-container), [[entrypoint: "entrypoint.sh" - docker compose](https://stackoverflow.com/questions/58438300/entrypoint-entrypoint-sh-docker-compose)](https://stackoverflow.com/questions/58438300/entrypoint-entrypoint-sh-docker-compose)

docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+if [ -f tmp/pids/server.pid ]; then
+  rm tmp/pids/server.pid
+fi
```
jakerenzella, line 6: This could just be:

`rm -f tmp/pids/server.pid`

jakerenzella:
Nice! This annoys me all the time

docker-entrypoint.sh
RayGuo-ergou: This @jakerenzella for you suggestion. It's done.

macite:
@RayGuo-ergou I dont think the **bundle exec** is needed in the _docker-entrypoint.sh_ as the CMD has this already. This will then also work if you want to run things outside of bundler.

RayGuo-ergou:
Hi @macite I have removed ` bundle exec ` in the *docker-entrypoint.sh* file.

jakerenzella:
Do we need the entry point file at all then?

RayGuo-ergou:
@jakerenzella Good point, maybe we can just add the ` rm -f  ` command into CMD?

jakerenzella:
> @jakerenzella Good point, maybe we can just add the ` rm -f  ` command into CMD?

Yep 👍 keep it simple

jakerenzella:
@RayGuo-ergou  this looks good but just need to double check that rm -f always returns a truthy value even if there's no file, otherwise the proceeding statements won't run

jakerenzella:
If not you can do rm -rf || command2 && command3 etc

RayGuo-ergou:
> @RayGuo-ergou this looks good but just need to double check that rm -f always returns a truthy value even if there's no file, otherwise the proceeding statements won't run

I tested to start the container that has no *server,pid* file and it's running okay. Is that what you mean, sorry I am not very sure.

jakerenzella:
> > @RayGuo-ergou this looks good but just need to double check that rm -f always returns a truthy value even if there's no file, otherwise the proceeding statements won't run
> 
> I tested to start the container when the _server,pid_ file is deleted and it's running okay. Is that what you mean, sorry I am not very sure.
> 
> `ls` can be excuted image

This looks all good then :)

PR #395, opened 2022-08-07
enhance: disable the x-runtime header in production

ublefo:
Disable the x-runtime header in production to (marginally) improve security. This is done by removing the Runtime middleware.

Tested in the development environment with the same directive.


macite:
@ublefo would this be useful to retain in the development environment?

ublefo:
Exactly, which is why it's only disabled in `production.rb`, not `development.rb`.

PR #296, opened 2020-10-03
NEW: Automate creation of portfolios at specified date

C-Blenco:
# Description

This feature ensures that in the event a student fails to compile their portfolio it will be generated at a certain date specified by unit administration (if they have a learning summary present).
Front-end implementation: doubtfire-lms/doubtfire-web#307

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Some tests have been written for the API, though more are currently being written (not yet commited).

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation if appropriate
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have created or extended unit tests to address my new additions
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

Draft pull request pending review for the current status of the feature.

PR #400, opened 2022-08-17
Feature: generate portfolios automatically after a configurable date

ublefo:
## Description

Implement the functionality to automatically generate the portfolio after a configurable date if the student have forgotten to do it.

Based on and supersedes: https://github.com/doubtfire-lms/doubtfire-api/pull/296

## Changes made

- [x] DB migrations for the new columns and retroactively set a sensible, valid autogen date for all existing units
- [x] Add new attributes to unit API code
- [x] Add ActiveRecord validation for the new date attribute
- [x] Update project API code to ensure the autogen flag is unset after user interaction
- [x] Introduce a new rake task to portfolio generation for qualified projects (see comments for details on the logic)
- [x] Add a new daily cron job to run the new rake task
- [x] Mail template changes to clearly indicate whether the portfolio is automatically generated or not
- [ ] Add unit tests for the introduced changes

## Testing required

- [x] Existing unit tests
- [x] Manual testing of portfolio generation in different situations
- [x] Testing the unit API changes
- [ ] New unit tests


app/api/units_api.rb
```diff
@@ -132,6 +134,14 @@ class UnitsApi < Grape::API
       end
     end
 
+    if unit_parameters[:portfolio_auto_generation_date]
```
macite, line 137: @ublefo can you move this into a validation of the unit that is triggered by any date change? e.g. change of start/end date or change of teaching period. Then this would trigger if you change any of those dates (though this may then cause challenges if you change the teaching period dates)

macite: https://guides.rubyonrails.org/active_record_validations.html

macite: See examples in unit and project

You can trigger this only when there are changes to a field:
```rb
class Project
    validate :tutorial_enrolment_same_campus, if: :will_save_change_to_enrolled?
end
```

And can also have custom if conditions
```rb
class Unit
   validates :code, uniqueness: { scope: :teaching_period, message: "%{value} already exists in this teaching period" }, if: :has_teaching_period?
end
```

app/api/units_api.rb
```diff
@@ -203,6 +215,21 @@ class UnitsApi < Grape::API
...
```
macite, line 218: You can then remove this as well

app/api/units_api.rb
```diff
@@ -132,6 +134,14 @@ class UnitsApi < Grape::API
...
+      # Ensure that portfolio generation date is within the teaching period
+      gen_date = unit_parameters[:portfolio_auto_generation_date]
+      unless gen_date > unit.start_date && gen_date <= unit.end_date
+        error!({ error: 'Portfolio auto generation date must be within the teaching period' }, 403)
+      end
+    end
+
     unit.update!(unit_parameters)
```
macite, line 145: Errors would then be raised on update!

app/api/units_api.rb
```diff
@@ -203,6 +215,21 @@ class UnitsApi < Grape::API
...
+        start_date = teaching_period.start_date
+        end_date = teaching_period.end_date
+      end
+
+      unless unit_parameters[:portfolio_auto_generation_date] > start_date && unit_parameters[:portfolio_auto_generation_date] <= end_date
+        error!({ error: 'Portfolio auto generation date must be within the unit start and end date' }, 403)
+      end
+    end
+
     unit = Unit.create!(unit_parameters)
```
macite, line 233: Errors would be raised on create!

app/views/portfolio_evidence_mailer/portfolio_failed.html.erb
```diff
@@ -47,6 +47,11 @@
   <main>
     <h1> <%= @doubtfire_product_name %> Notification </h1>
     <p>Hi <%= @student.first_name %>,</p>
+    <% if @project.portfolio_auto_generated %>
+      <p>
+        Profile generation has been automatically triggered because you have not yet generated it beyond the date that the unit chair has specified.
```
macite: We can probably make this more friendly... "Profile generation has been automatically triggered. You can customise your portfolio and recreate up until the due date indicated by the teaching staff."

app/api/units_api.rb
macite, line 137: Move this to an active record validation - https://guides.rubyonrails.org/active_record_validations.html

See examples in Project and Unit. eg

```rb
# Unit
validates :code, uniqueness: { scope: :teaching_period, message: "%{value} already exists in this teaching period" }, if: :has_teaching_period?
# Project
validate :tutorial_enrolment_same_campus, if: :will_save_change_to_enrolled?
```

You can then remove these checks and they will trigger on Update! and Create!

app/views/portfolio_evidence_mailer/portfolio_failed.text.erb
```diff
@@ -1,5 +1,7 @@
 Hi <%= @student.first_name %>,
 
+<% if @project.portfolio_auto_generated %>Profile generation has been automatically triggered because you have not yet generated it beyond the date that the unit chair has specified.<% end %>
```
macite, line 3:
```suggestion
<% if @project.portfolio_auto_generated %>Profile generation has been automatically triggered. You can customise your portfolio and recreate up until the due date indicated by the teaching staff.<% end %>
```

app/views/portfolio_evidence_mailer/portfolio_ready.html.erb
```diff
@@ -47,6 +47,11 @@
   <main>
     <h1> <%= @doubtfire_product_name %> Notification </h1>
     <p>Hi <%= @student.first_name %>,</p>
+    <% if @project.portfolio_auto_generated %>
+      <p>
+        Profile generation has been automatically triggered because you have not yet generated it beyond the date that the unit chair has specified.
```
macite, line 52:
```suggestion
        Profile generation has been automatically triggered. You can customise your portfolio and recreate up until the due date indicated by the teaching staff.
```

app/views/portfolio_evidence_mailer/portfolio_ready.text.erb
```diff
@@ -1,5 +1,7 @@
 Hi <%= @student.first_name %>,
 
+<% if @project.portfolio_auto_generated %>Profile generation has been automatically triggered because you have not yet generated it beyond the date that the unit chair has specified.<% end %>
```
macite, line 3:
```suggestion
<% if @project.portfolio_auto_generated %>Profile generation has been automatically triggered. You can customise your portfolio and recreate up until the due date indicated by the teaching staff.<% end %>
```

db/migrate/20220817123038_portfolio_automation.rb
```diff
@@ -0,0 +1,6 @@
+class PortfolioAutomation < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :portfolio_auto_generated, :boolean, default: false, null: false
+    add_column :units, :portfolio_auto_generation_date, :datetime
```
macite, line 4: Do we default this to 2 weeks after the draft learning summary report for the existing units that have it?

macite: And the unit is not over ... so now < unit end date

lib/tasks/generate_pdfs.rake
```diff
@@ -55,6 +55,24 @@ namespace :submission do
...
+        # - current date is after configured automatic generation dates for the associated unit
+        # - portfolio compilation has not been requested already
+        # - portfolio is still not available
+        # Additionally, set the portfolio_auto_generated property to true to indicate this is auto-generated
+        Project.joins(:unit)
+               .where(units: { active: true })
+               .where("units.portfolio_auto_generation_date < ?", Date.today)
+               .where(compile_portfolio: false)
+               .reject(&:portfolio_available)
+               .each do |project|
```
macite, line 70: can we use `update_all` here to just run a single update statement on the database?

https://guides.rubyonrails.org/active_record_basics.html#update

macite (changes requested):
Looking very close, just a few things we can improve to make this work better.

lib/tasks/generate_pdfs.rake
ublefo: `portfolio_available` is not a db query, it's a method of the `Project` class. Since this check is mandatory and we don't really want another boolean in the db that can get desynchronized with the filesystem state, we can't use `update_all` here. I am aware this is a performance hit, and I'll be moving this to a new rake task and create a new daily cron job to run it for now to reduce the overhead.

app/api/units_api.rb
ublefo: Yes, all the validation stuff from the old PR will be removed and converted to ActiveRecord validations. The "is date within teaching period" check can probably be ignored since we are already validating the start/end dates if a teaching period is defined.

db/migrate/20220817123038_portfolio_automation.rb
ublefo: > Do we default this to 2 weeks after the draft learning summary report for the existing units that have it?

What about units that doesn't have it? Since this is a failsafe, how about we just default it to 3 days before unit end date to make it simpler?

> And the unit is not over
I would rather not have this as null if possible, end date - 3 should be a safe value to apply to all units, including inactive ones.

macite (approved):
Great job in recreating this...

PR #406, opened 2022-09-09
fix: hard-code date in newpax to avoid nil dates

ublefo:
Students have been reporting higher error rates after newpax has been introduced due to the fact that it can't cope with nonexistent creation date in the PDF metadata. Fix this by hard-coding the date for now as an emergency fix, and we will look into addressing this properly later on.


ublefo:
I've reported the issue to upstream and it has been fixed properly. A new version of the package has been published to CTAN as well, which should be available for install in the mirrors in the upcoming days.

We will need to remove the patch entirely after the new version is available in the mirrors, because the package version was not incremented in the new release and my patch logic was checking the package version.

PR #434, opened 2024-03-27
CI fix

TinyServal:
Various CI related fixes and improvements:
1. Implemented container caching to bring turnaround time from [52 minutes](https://github.com/doubtfire-lms/doubtfire-api/actions/runs/8396126610/job/22996854649) to [16 minutes](https://github.com/ublefo/doubtfire-api/actions/runs/8439026003/job/23112641848) after the container is built successfully.
2. Set up redis in the workflow and make sure sidekiq is configured to use the correct URL from the environment if it's present.
3. Upgrade all workflows to the latest versions.
4. Install `ruby-lsp` in the development environment for the corresponding vscode extension, which is currently already installed in the devcontainer but completely useless due to the gem being missing.
5. Implement better error reporting in `database_populator.rb` to report the error messages when there's something wrong with the test data. Troubleshooting test db initialization issues should be much easier now.
6. Documented the two new redis related env variables.
7. Skip unit tests for documentation updates.
8. Created a dependabot config to check for github actions workflow updates weekly.


maddernd:
Approved by DM, just waiting on @macite approval.

PR #438, opened 2024-04-22
fix: task file import

TinyServal:
Fixes task file import bug where a task file import will attach the task sheets and resources to the wrong task. Now we use the longest match to ensure the task file is imported to the correct task.

A new task `T10` has been added to test this behavior, and now the unit test will validate the content of the task sheets for `T1` and `T10` so we can catch this in the future.


app/models/unit.rb
```diff
@@ -1634,7 +1634,8 @@ def import_task_files_from_zip(zip_file)
         if (File.extname(file.name) == '.pdf') || (File.extname(file.name) == '.zip')
           found = false
           task_definitions.each do |td|
-            next unless /^#{td.abbreviation}/ =~ file_name
+            # strip extensions and look for exact match
+            next unless td.abbreviation == File.basename(file.name, ".*")
```
macite: I dont want to force exact match - checking the start should be ok but we need to search for the longest match and use that.

macite (changes requested):
Fix file matching to allow partial match of filename, searching for the longest matching abbreviation

app/models/unit.rb
TinyServal: Done

PR #439, opened 2024-04-27
fix: truncate long lines in PDF conversion

TinyServal:
This fixes the PDF processing issue when there are super long lines in the submitted code files by using `cut` and `fold` from coreutils to wrap lines. Additionally, introduce https://github.com/ruby/shellwords for proper shell escape.

There are two limits configured, the hard limit is 1000 characters, and the soft limit is 160 characters. Any lines over 1000 characters will be truncated, and then line breaks will be applied for any lines over 160 characters long. If the rendered file has been modified, a warning will be added into the PDF document to indicate the rendered file differs from the original submission. Unit tests and test files have been added for these changes.

Inkscape has been replaced with `librsvg` for SVG to PDF conversion in jupynotex. Inkscape pulls in a lot of unnecessary dependencies (dbus, gtk, etc.), and it emits annoying irrelevant errors that has to be redirected to `/dev/null`. `rsvg-convert` is much more lightweight in comparison, performs equally well and has none of the downsides of Inkscape: https://gitlab.gnome.org/GNOME/librsvg

Generated result (excluding Jupyter Notebook):
image

Jupyter Notebook:
image


app/models/task.rb
```diff
@@ -994,13 +994,36 @@ def init(task, is_retry)
...
-      logger.debug "Running QPDF on all documents before rendering to repair any potential broken files."
+      logger.debug "Rendering PDF for a task, preprocessing attachments."
       @files.each do |f|
         if f[:type] == "document"
+          logger.debug "Running QPDF on #{f[:path]} before rendering to repair any potential broken files."
           FileHelper.qpdf(f[:path])
+        elsif f[:type] == "code"
+          extension = File.extname(f[:path])[1..-1]
+          case extension
+          when "ipynb"
```
macite: Why not ipynb? I think these may be more likely to have the embedded data sets.

app/helpers/file_helper.rb
```diff
@@ -576,6 +578,39 @@ def task_submission_identifier_path_with_timestamp(type, task, timestamp)
     "#{task_submission_identifier_path(type, task)}/#{timestamp.to_s}"
   end
 
+  # Apply line wrapping to a given file.
+  # The default 160-character limit is about two lines in the rendered PDF.
+  # There is also a hard limit of 1000 characters, anything longer will be truncated to 1000 characters.
+  def line_wrap(path, width: 160)
```
macite, line 584: Why create an extra tmp file that has to be removed later? Why no re-write this file if needed? If this only occurs during rendering then it is working on a copy already - so there is no need to preserve the original.

app/models/task.rb
```diff
@@ -994,13 +994,36 @@ def init(task, is_retry)
...
       end
+      logger.debug "Preprocessing complete, rendering file."
       render_to_string(template: '/task/task_pdf', layout: true)
+    ensure
+      logger.debug "Cleaning up line-wrapped temporary files after rendering."
+      @files.each do |f|
+        next unless f[:type] == "code"
+        name = File.basename(f[:path])
+        dir = File.dirname(f[:path])
+        tempfile_path = File.join(dir, "#{name}.tmp")
```
macite: Rewriting the original avoids this as well.

app/views/task/task_pdf.pdf.erb
```diff
@@ -112,8 +112,20 @@ No Tutor % Supervisor's Name
...
+      # for code files (except Jupyter Notebook), always use files with the .tmp suffix
+      # if the file has not been modified, it will be a symlink instead
+      # Jupyter Notebooks are not processed with line wraps
+      dir = File.dirname(file[:path])
+      name = File.basename(file[:path])
+      pygments_lang = Task.pygments_lang(File.extname(file[:path])[1..-1])
+      wrapped_file = File.join(dir, "#{name}.tmp")
   %>
-\inputminted[breaklines,linenos,breakanywhere,tabsize=4]{<%= Task.pygments_lang(File.extname(file[:path])[1..-1]) %>}{<%= file[:path] %>}
+    <% unless File.symlink?(wrapped_file) %>
```
macite: If we re-wrote the original we would have to think of a different way to handle this...

macite (changes requested):
Can you see if we can do this with re-writing the original rather than creating the tempfile?

app/models/task.rb
TinyServal: Modifying ipynb without parsing it properly will break the notbook entirely, and jupynotex will fail to process it. The hope is that jupynotex will be able to handle it but I'm not 100% confident that is the case, will have to add another test file to make sure it works properly.

TinyServal: Fixed

PR #440, opened 2024-04-30
refactor: change extension granularity from weeks to days

TinyServal:
This has been a highly requested feature, providing better granularity for extensions improves flexibility for both the students and staff.

Summary of the changes:
- New db migration to rename the db columns
- Adjusted all extension related logic to use days instead of weeks
- Adjusted all unit tests to reflect the change

Note: the "grant maximum number of allowed extensions when requested extension is longer than the limit" behavior has not been removed in this refactor, so if a student requests 7 days and the due date is 2 days away, only 2 days will be granted just like before. The removal of this behavior can be done at a later date.

Frontend changes: https://github.com/doubtfire-lms/doubtfire-web/pull/830


TinyServal:
Closing due to lack of interest from the maintainers.

PR #445, opened 2024-09-08
Feat: Numbas updated with polymorphic association

satikaj:
# Description

Latest commit introduces polymorphic association to task comments. Currently, task comments can belong to either Scorm test attempts or Overseer assessments.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Unit tests passed. Comments display as usual in the frontend.

# Checklist:


db/migrate/20240601103707_add_polymorphic_association_to_comment.rb
```diff
@@ -0,0 +1,16 @@
+class AddPolymorphicAssociationToComment < ActiveRecord::Migration[7.1]
+  def change
+    remove_index :task_comments, :overseer_assessment_id
+
+    add_column :task_comments, :commentable_type, :string
+    rename_column :task_comments, :overseer_assessment_id, :commentable_id
+
+    TaskComment.find_each do |comment|
+      if comment.commentable_id.present?
```
macite: You should be able to do this with a where - TaskComment.where('NOT commendable_id IS NULL').find_each

db/migrate/20240601103707_add_polymorphic_association_to_comment.rb
```diff
@@ -0,0 +1,16 @@
...
+        comment.update(commentable_type: 'OverseerAssessment')
```
macite: I think you could then do update all... something like this:

```suggestion
     TaskComment.where('NOT commendable_id IS NULL').in_batches.update_all(commentable_type: 'OverseerAssessment')
```

macite (changes requested):
Looking super close - the Task Comment table is huge, so I think we can optimise the update... see the comment. @satikaj do you have time to make these changes?

db/migrate/20240601103707_add_polymorphic_association_to_comment.rb
satikaj: I've made the changes now, Andrew

PR #446, opened 2024-09-10
Refactor: LaTeX processing in external Docker container

b0ink:
**Depends on https://github.com/doubtfire-lms/doubtfire-deploy/pull/29**

# Description

This update moves LaTeX processing into a dedicated TeX Live container. Instead of running `lualatex` from the [latex_config](https://github.com/thoth-tech/doubtfire-api/pull/38/files#diff-167b948375f6e2b72c3cc40acc447e3403e40953443451e7bd1501fd646a9309), the [latex_run.sh](https://github.com/b0ink/doubtfire-api/blob/ce988b26de5b18a919ff980f33cf08aff584bbeb/lib/shell/latex_run.sh) script is executed after being copied into the working directory (along with other input files such as `input.tex` and `jupynotex.py`).

`latex_run.sh` then runs `docker exec ...` into the TeX Live container and executes [latex_build.sh](https://github.com/b0ink/doubtfire-api/blob/ce988b26de5b18a919ff980f33cf08aff584bbeb/lib/shell/latex_build.sh), which is in charge of running the `lualatex` commands.

`latex_run.sh` relies on two environment variables to be set:
```env
LATEX_CONTAINER_NAME=formatif-texlive-container # Name of the TeX Live container
LATEX_BUILD_PATH=/texlive/shell/latex_build.sh # Path on the TeX Live container
```

The `render_to_string()` method that renders the PDFs is now wrapped in a `generate_pdf()` method defined in [latex_helper.rb](https://github.com/b0ink/doubtfire-api/blob/ce988b26de5b18a919ff980f33cf08aff584bbeb/app/helpers/latex_helper.rb), ensuring the environment variables are set before executing `latex_run.sh` via `render_to_string`.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- Use the modified docker-compose from https://github.com/doubtfire-lms/doubtfire-deploy/pull/29
- If you encounter issues with the setup, remove any previous Docker devcontainer builds and volumes.
- Run `rake test` in `doubtfire-api` to validate PDF generation.
- Optionally, run `rake db:simulate_signoff` which will create several tasks and render them with lualatex.
- You can also manually create a portfolio when logged in as a student, upload different test files, click "Create my portfolio", then run `doubtfire-api/lib/shell/generate_pdfs.sh`

# Checklist:


macite:
Hi @b0ink, it looks like the merge commit [fb0bc26](https://github.com/doubtfire-lms/doubtfire-api/pull/446/commits/fb0bc2657d62733edbfd6192755304e24d669e4d) has not merged the change log well. Can you recreate the PR and merge into the new/scorm branch? You could just revert the last two merge commits, and then force push to your branch. That may update this PR...

macite:
We also probably need to make sure to update the GitHub actions so that it tests using the separate containers

docker-compose.yml
```diff
@@ -71,3 +76,17 @@ services:
...
+      context: .
+      dockerfile: texlive.Dockerfile
+    volumes:
+      - ../data/student-work:/student-work
+      - ./public/assets/images:/doubtfire/public/assets/images
+      - ./test_files:/doubtfire/test_files
+      - ./tmp/rails-latex:/workdir/texlive-latex
+    depends_on:
+      - df-api
+    command: /bin/sh -c "while sleep 5000; do :; done"
```
macite: You should be able to use `while sleep infinity; ...` here

app/models/task.rb
```diff
@@ -1087,7 +1089,7 @@ def make_pdf
         end
       end
       logger.debug "Preprocessing complete, rendering file."
-      render_to_string(template: '/task/task_pdf', layout: true)
+      generate_pdf(template: "/task/task_pdf")
```
macite: Use ' around text - " if the text contains string interpolation.

app/helpers/latex_helper.rb
```diff
@@ -0,0 +1,7 @@
+module LatexHelper
+  def generate_pdf(template:)
+    raise "LATEX_CONTAINER_NAME is not set" if ENV['LATEX_CONTAINER_NAME'].nil?
```
macite: Use ' around text - " if the text contains string interpolation or escape characters. Using ' will be faster where these do not exist.

app/helpers/latex_helper.rb
```diff
@@ -0,0 +1,7 @@
...
+    raise "LATEX_BUILD_PATH is not set" if ENV['LATEX_BUILD_PATH'].nil?
```
macite: ' instead of "

app/models/pdf_generation/project_compile_portfolio_module.rb
```diff
@@ -66,8 +68,10 @@ def init(project, is_retry)
       end
 
       def make_pdf
-        render_to_string(template: '/portfolio/portfolio_pdf', layout: true)
+        logger.debug "Running make_pdf: (portfolio)"
```
macite: ' instead of "

app/models/pdf_generation/project_compile_portfolio_module.rb
```diff
@@ -66,8 +68,10 @@ def init(project, is_retry)
...
+        generate_pdf(template: "/portfolio/portfolio_pdf")
```
macite: ' instead of "

texlive.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM texlive/texlive:latest
+
+RUN apt-get update \
+    && apt-get install -y \
+    imagemagick \
+    inkscape \
+    librsvg2-bin \
```
macite, line 7: We should be able to remove inkscape and librsvg2-bin from the other Dockerfiles as I think these are only used in the pdf creation.

texlive.Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY ./lib/shell/latex_build.sh /texlive/shell/latex_build.sh
+RUN chmod +x /texlive/shell/latex_build.sh
+
+CMD ["sh", "-c", "while sleep 5000; do :; done"]
```
macite: infinity should work here too

lib/shell/latex_run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# This script is copied into tmp/rails-latex/$WORK_DIR/ and executed by rails-latex
+
+WORK_DIR=$(basename "$PWD")
+docker exec -t $LATEX_CONTAINER_NAME $LATEX_BUILD_PATH $WORK_DIR
```
macite, line 5: We should check that these environment variables exist - either here or in the application.rb file. Then report an error so that the config can be checked.
$LATEX_CONTAINER_NAME 
$LATEX_BUILD_PATH

macite: Should we default LATEX_BUILD_PATH to /texlive/shell/latex_build.sh if it is not set?

macite (changes requested):
Looks pretty close now - just a few minor things to look at.

PR #447, opened 2024-09-20
Feature/jplag implementation

JackSCarroll:
Depends on [ doubtfire-lms/doubtfire-deploy#28 ](https://github.com/doubtfire-lms/doubtfire-deploy/pull/28)
Depends on [ doubtfire-lms/doubtfire-web#873 ](https://github.com/doubtfire-lms/doubtfire-web/pull/873)
Feasibility document: [ thoth-tech/documentation#528 ](https://github.com/thoth-tech/documentation/pull/528)

# Description

This branch adds JPLAG software plagiarism detection to OnTrack. The `unit_similarity_module.rb` file has been modified to extend the `check_similarity function` to perform a docker exec command to generate a jplag report for each task with "done" files in a unit.
It extracts the code files from the "done" zip files and temporarily stores them in `tmp/jplag/unitcode/task` while the jplag report is being generated.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

- Use the modified docker-compose and added jplag.Dockerfile from [ Feature/add jplag plagiarism detection #28 ](https://github.com/doubtfire-lms/doubtfire-deploy/pull/28)
- Create a new task in a unit and have at least 2 student users upload similar code files for the task
- Change task due date to be in the past, so submissions can be accepted in the check_similarity check
- Run `rake submission:generate_pdfs` while cd'ed in doubtfire-api. This will create the "done" files.
- Run `rake submission:check_plagiarism` while cd'ed in doubtfire-api. This will run jplag on all units and tasks.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation if appropriate
- [X] My changes generate no new warnings


macite:
Can you recreate commit 90ef645edaaf057e90f567aa9409ad72a37afb89 with a chore tag at the start of the commit message.

app/models/similarity/unit_similarity_module.rb
```diff
@@ -14,17 +14,26 @@ def last_plagarism_scan
...
+
     begin
       logger.info "Checking plagiarsm for unit #{code} - #{name} (id=#{id})"
 
       task_definitions.each do |td|
+        # making temp directory for each task - jplag
+        tasks_dir = root_work_dir.join(td.id.to_s)
+        FileUtils.mkdir_p(tasks_dir)
+
         next if td.moss_language.nil? || td.upload_requirements.nil? || td.upload_requirements.select { |upreq| upreq['type'] == 'code' && upreq['tii_check'] }.empty?
```
macite: should this still be moss_language?

app/models/similarity/unit_similarity_module.rb
```diff
@@ -35,6 +44,7 @@ def check_moss_similarity(force: false)
         tasks = tasks_for_definition(td)
         tasks_with_files = tasks.select(&:has_pdf)
 
+
```
macite, line 49: Remove unnecessary empty lines

macite: Have to think if this is worth changing...

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+      # only check code files marked for similarity checks
+      next unless upreq['type'] == 'code' && upreq['tii_check']
+
+      pattern = task_definition.glob_for_upload_requirement(idx)
+
+      tasks_with_files.each do |t|
+        t.extract_file_from_done(tasks_dir, pattern, ->(_task, to_path, name) { File.join(to_path.to_s, t.student.username.to_s, name.to_s) })
+      end
+      puts "Starting JPLAG container to run on #{tasks_dir}"
+      tasks_dir_split = tasks_dir.to_s.split('/workspace/doubtfire-api')[1]
```
macite: hardcoded dev container path... won't work in production

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+    task_definition.upload_requirements.each_with_index do |upreq, idx|
+      # only check code files marked for similarity checks
+      next unless upreq['type'] == 'code' && upreq['tii_check']
+
+      pattern = task_definition.glob_for_upload_requirement(idx)
+
+      tasks_with_files.each do |t|
+        t.extract_file_from_done(tasks_dir, pattern, ->(_task, to_path, name) { File.join(to_path.to_s, t.student.username.to_s, name.to_s) })
+      end
+      puts "Starting JPLAG container to run on #{tasks_dir}"
```
macite: Use logging not puts

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+      tasks_dir_split = tasks_dir.to_s.split('/workspace/doubtfire-api')[1]
+
+      # Check if the directory exists and create it if it doesn't
+      results_dir = "/jplag/results/#{unit_code}"
+      `sudo docker exec jplag sh -c 'if [ ! -d "#{results_dir}" ]; then mkdir -p "#{results_dir}"; fi'`
```
macite: sudo?

macite: Should this be outside the upload requirements loop? What happens if you have multiple files being checked in the same task definition?

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+
+      # Remove existing result file if it exists
+      result_file = "#{results_dir}/#{task_definition.id}-result.zip"
+      `sudo docker exec jplag sh -c 'if [ -f "#{result_file}" ]; then rm "#{result_file}"; fi'`
+
+      case type_data[1]
```
macite: Is this all the languages available?

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+      when 'csharp'
+        file_lang = 'csharp'
+      when 'cc'
+        file_lang = 'cpp'
+      end
+
+      # Run JPLAG
+      puts "THE FILE TYPE IN MOSS: #{type_data[1]}"
```
macite: puts

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+      `sudo docker exec jplag java -jar /jplag/myJplag.jar #{tasks_dir_split} -l #{file_lang} --similarity-threshold=0.8 -M RUN -r #{results_dir}/#{task_definition.id}-result`
+    end
+
+    # Delete the extracted code files from tmp
+    tmp_dir = Rails.root.join("tmp", "jplag")
+    puts "Deleting files in: #{tmp_dir}"
```
macite: puts

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+    puts "Files to delete: #{Dir.glob("#{tmp_dir}/*")}"
```
macite: puts

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,51 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+    FileUtils.rm_rf(Dir.glob("#{tmp_dir}/*"))
+
+    self
```
macite, line 131: Where are the similarity matches created?

macite: We need to parse the report and link related tasks

macite (changes requested):
This looks like a good start, but there seems to be quite a bit to do still.

app/models/similarity/unit_similarity_module.rb
JackSCarroll: Because some of the language names in MOSS and JPLAG differ, i've added a check for c++ and python.
Currently only supporting languages defined from the task definition ui `task-definition-upload.component.html`
Have added comments to both `task-definition-upload.component.html` and `unit-similarity-module.rb` as reminder to check MOSS and JPLAG language defintions

app/models/similarity/unit_similarity_module.rb
JackSCarroll: Sudo is needed to make the `docker exec` commands work. Not entirely sure why 🙃.
Have moved the dir checks outside the loop and still works as expected.

app/models/similarity/unit_similarity_module.rb
JackSCarroll: Have changed to Rails.root

app/models/similarity/unit_similarity_module.rb
JackSCarroll: I'm using the upload requirement language from the task definition ui `task-definition-upload.component.html`.
Feel like it makes sense to use existing language definition considering the task creator has already decided if plagiarism checks are necessary from MOSS.
So if no requirement for MOSS, no requirement for JPLAG. 
Let me know if you don't think is this appropriate.

macite: With this we should remove the old moss code, so these could be renamed...

app/models/similarity/unit_similarity_module.rb
```diff
@@ -222,4 +235,60 @@ def add_done_files_for_plagiarism_check_of(task_definition, tmp_path, to_check,
...
+
+      tasks_with_files.each do |t|
+        t.extract_file_from_done(tasks_dir, pattern, ->(_task, to_path, name) { File.join(to_path.to_s, t.student.username.to_s, name.to_s) })
+      end
+
+      logger.info "Starting JPLAG container to run on #{tasks_dir}"
+      root_dir = Rails.root.to_s
+      tasks_dir_split = tasks_dir.to_s.split(root_dir)[1]
+
+      # Set the file language based on the type
```
macite: We can be removing MOSS with this - so the languages can be updated to those we want from JPlag. This would be part of the front end change as well.

macite (changes requested):
Looks like we need to remove the Moss code, switch the languages in the task definitions, and add in some unit tests.

app/models/similarity/unit_similarity_module.rb
b0ink: `docker exec` doesn't require `sudo` in production since it'll run as the root user, but the permission errors in the devcontainer are due to it using the `vscode` user

You can fix this by [setting permissions on `/var/run/docker.sock`](https://github.com/doubtfire-lms/doubtfire-deploy/pull/29/commits/a3af75f200fdb4be9ce5afc621c906606bbbeb73#diff-79495a5ff44c04cd52db3561d48e92865b1b47e8738ddc27a489224de4914daaR10) in https://github.com/doubtfire-lms/doubtfire-deploy

app/models/similarity/unit_similarity_module.rb
JackSCarroll: I've added a function to create the similarity matches. [Here's](https://github.com/doubtfire-lms/doubtfire-api/pull/447/commits/2a237597b29b49a833fd55e6bcc4a2e9a604b8d0) the commit.
It opens the overview.json file from the report and extracts the max similarity between 2 submissions.
It also finds the task id by looking in /files/student_name/task_id and grabbing the task id folder name; it matches the student names from the two that are being looked at currently in overview.json.

macite:
Merged into 10.0.x

PR #451, opened 2024-12-17
Fix: no auth method mailer

jakerenzella:
## WARNING: This change may require servers to set `DF_SMTP_AUTHENTICATION` to `plain` as that was the previous default.

Currently, mailer auth defaults to `plain`.

Some mailers don't rely on mailer auth, and instead use internal IP ranges.

No-auth method smtp relies on the lack of an auth key, and there is no valid value to default to.

The defaul plain method meant that non-auth mailers can't set the auth method correctly.

## WARNING: This may require servers to set `DF_SMTP_AUTHENTICATION` to `plain` as that was the previous default.


macite:
Implemented through no_auth / none setting.

PR #455, opened 2025-01-26
enhance: data pagination

nodogx:
# Description
## Added Data Pagination

Now, instead of loading all the data histories at once, the backend just gives the data based on the webpage and a limited number or data. For example, If there are 30 records, and if the page 1 has 10 records, this will only give the 10 records in the backend with the total number of records (needed for page count).  

When the user goes to page 2 for example, the next set of data will be loaded. This way, if there are 500 histories for example, everything will not be loaded at once. 

This will help boost webpage load times

Screenshot 2024-12-23 at 2 29 02 am

# How Has This Been Tested?

Login as any user
Go to the dashboard (where all the target history is been shown)
Go to the inspect page in the browser.
Go to the Network tab, and see how the data is loaded every time when the user is navigated to a new page

## Testing Checklist:

- [x] Tested in latest Chrome
- [x] Tested in latest Safari
- [x] Tested in latest Firefox

# Checklist:

- [x] I have commented my code in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have requested a review from @macite and @jakerenzella on the Pull Request

Front end - https://github.com/doubtfire-lms/doubtfire-web/pull/913


b0ink:
We'll probably want to use a more global framework such as https://github.com/bsm/grape-kaminari for pagination, and closely integrate it with out entity service manager

PR #521, opened 2025-09-29
feat: pdf diff generation

b0ink:


b0ink:
not feasible in its current state

PR #566, opened 2026-02-01
Integrate Approved Session Tracking Fixes into 10.0.x

rammakablecode:
### References:
Approved PR: https://github.com/thoth-tech/doubtfire-api/pull/83
Unmerged PR: https://github.com/thoth-tech/doubtfire-api/pull/74

# Description

This PR merges the changes from the approved PR [#83](https://github.com/thoth-tech/doubtfire-api/pull/83) into the 10.0.x branch. It ensures the session tracking logic integrated includes schema fixes, environment blockers, and backend improvements previously approved in [#83](https://github.com/thoth-tech/doubtfire-api/pull/83).

This PR directly references the approved work from [#83](https://github.com/thoth-tech/doubtfire-api/pull/83) and makes it available in the target branch for further development and frontend integration.

#### Fixes identified issues that blocked development and limited the usefulness of the existing schema:
- MySQL tablespace errors (errno: 194) that prevented successful migrations in Windows-hosted Docker environments.
- CRLF line endings in shell scripts causing Dev Container startup failures.
- Inefficient frontend computation of session_activity gaps.

Frontend impact
Persisting duration_minutes enables direct aggregation via the API and removes the need for complex time-delta calculations in the Angular frontend. This change is intended to support(https://github.com/thoth-tech/doubtfire-web/pull/402). By providing this data via the API, we avoid complex time-delta calculations in the Angular frontend, allowing 402 to be refactored to connect directly to this backend logic since the frontend contributions are yet to connect to exisitng backend.

## Type of change

# How Has This Been Tested?

- [x] Test A: Migration Sequence
Executed bundle exec rails db:migrate:redo VERSION=20250922103033.

   Verified:
   `marking_sessions` table is created successfully
   `duration_minutes` column exists
   No MySQL tablespace errors occur

- [x] Test B: Session Tracking Logic
Ran `bundle exec ruby -I test test/services/session_tracker_test.rb.`
   Result: 2 runs, 13 assertions, 0 failures, 0 errors.
   Ran `bundle exec rails test test/api/marking_sessions_api_test.rb`
   6 runs, 67 assertions, 0 failures, 0 errors, 0 skips

image

- 
Screenshot 2026-01-29 162100

   Confirmed:
   15-minute "sticky" session window works 
   `duration_minutes` is correctly updated upon activity.

# Checklist:

### DIFF
--- a/app/models/marking_session.rb
+++ b/app/models/marking_session.rb
@@ -15,6 +15,14 @@ class MarkingSession < ApplicationRecord
 
   def update_session_details
     now = DateTime.now
-    update(end_time: now)
+    if start_time.present?
+      duration = ((now.to_f - start_time.to_f) / 60).to_i
+      update(
+        end_time: now,
+        duration_minutes: duration
+      )
+    else
+      update(end_time: now, duration_minutes: 0)
+    end
   end
 end
--- a/app/services/session_tracker.rb
+++ b/app/services/session_tracker.rb
@@ -18,10 +18,10 @@ class SessionTracker
       project_id: project.id,
       task_id: task&.id,
       task_definition_id: task&.task_definition_id,
-      created_at: Time.zone.now
+      created_at: DateTime.now
     )
 
-    session.update_session_details
+    session.update_session_details if action == 'assessing'
 
     activity
   end
--- a/db/migrate/20250922103033_create_marking_sessions.rb
+++ b/db/migrate/20250922103033_create_marking_sessions.rb
@@ -6,6 +6,7 @@ class CreateMarkingSessions < ActiveRecord::Migration[7.1]
       t.string :ip_address
       t.datetime :start_time
       t.datetime :end_time
+      t.integer :duration_minutes, default: 0
       t.boolean :during_tutorial
 
       t.timestamps
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -191,6 +191,7 @@ ActiveRecord::Schema[8.0].define(version: 2025_11_02_221253) do
     t.string "ip_address"
     t.datetime "start_time"
     t.datetime "end_time"
+    t.integer "duration_minutes", default: 0
     t.boolean "during_tutorial"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
--- a/test/api/marking_sessions_api_test.rb
+++ b/test/api/marking_sessions_api_test.rb
@@ -150,16 +150,15 @@ class MarkingSessionsApiTest < ActiveSupport::TestCase
     assert_not last_activity.nil?
 
     # > Force the session to be 10 minutes in the past
-    travel 10.minutes
-    travel 5.seconds # Ensure we're past the 10 minute mark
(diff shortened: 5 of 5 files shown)


db/migrate/20250922103033_create_marking_sessions.rb
```diff
@@ -6,6 +6,7 @@ def change
       t.string :ip_address
       t.datetime :start_time
       t.datetime :end_time
+      t.integer :duration_minutes, default: 0
```
b0ink, line 9: You'll need to start a new migration if you want to make changes to the schema, as this is already shipped in production

app/services/session_tracker.rb
```diff
@@ -18,10 +18,10 @@ def self.record_assessment_activity(action:, user:, project:, ip_address:, task:
       project_id: project.id,
       task_id: task&.id,
       task_definition_id: task&.task_definition_id,
-      created_at: Time.zone.now
+      created_at: DateTime.now
     )
 
-    session.update_session_details
+    session.update_session_details if action == 'assessing'
```
b0ink, line 24: We want various actions to let the backend know the tutor is still active - this includes adding comments, opening different submissions, etc.

We can't limit this to just the `assessing` action, because a tutor could - for example - spend 2 hours opening and reviewing submissions, leaving comments, and not change the status of any tasks (triggering the `assessing` action) - this would result in multiple sessions being created back-to-back once the previous session is older than 15 minutes.

app/models/marking_session.rb
```diff
@@ -15,6 +15,14 @@ def duration_minutes
...
-    update(end_time: now)
+    if start_time.present?
+      duration = ((now.to_f - start_time.to_f) / 60).to_i
+      update(
+        end_time: now,
+        duration_minutes: duration
+      )
+    else
+      update(end_time: now, duration_minutes: 0)
+    end
```
b0ink, line 26: duration_minutes is redundant and should not be part of the schema.
The cost of calculating duration from start_time and end_time is negligible, and storing it would duplicate derived data.

Persisting duration_minutes introduces a risk of data inconsistency - for example, if one value is updated without the other - leading to conflicting sources of truth

We [dynamically calculate](https://github.com/doubtfire-lms/doubtfire-api/blob/339b6493ec73432ef74e96c9983fb17d0659de92/app/models/marking_session.rb#L11-L14) duration_minutes when we expose the MarkingSession entity

app/services/session_tracker.rb
```diff
@@ -18,10 +18,10 @@ def self.record_assessment_activity(action:, user:, project:, ip_address:, task:
       project_id: project.id,
       task_id: task&.id,
       task_definition_id: task&.task_definition_id,
-      created_at: Time.zone.now
+      created_at: DateTime.now
```
b0ink, line 21: Time.zone.now should be used

b0ink (commented):
Thanks for your contributions, but I don't think we require any of these changes and I have added comments explaining why

If you were looking to optimise how we handle the session duration, it would probably be even better not to expose the `duration_minutes` from the backend at all, reducing the size of the data sent from the server, and to do the calculation client-side the performance cost is negligible

PR #568, opened 2026-02-10
feat: clang-query based concept usage checker

WhyPenguins:
**Note** - this PR isn't necessarily intended to be merged directly :) I'm unfamiliar with this project, and simply tried to guess at the relevant locations for these scripts. They will probably need to be relocated. They are also not in the language of the project - since they are relatively standalone I hope this is alright :)

# Description

This PR adds a script `check_concept_usage.py` that uses `clang-query` to check if a C++ file matches any rules from a named rule set. It can be called as follows:
```
python check_concept_usage.py "submission.cpp" chapter_05_structuring_code desc
```
and it will output a list of one-line descriptions for each rule that was matched. 

If it fails to run to completion, it will output errors on stderr, and leave stdout empty.

The rules are defined in `rules.json` - here's an example of the layout:
```
{
...
  "chapter_04_control_flow": {
    "includes": ["chapter_05_structuring_code"],
    "rules": [
      {
        "queries": ["functionDecl(isExpansionInMainFile(), unless(hasName(\"main\")), unless(cxxMethodDecl()))"],
        "id": "Function",
        "description": "Check that all your code is in the one function for now."
      },
      {
        "queries": ["cxxThrowExpr(isExpansionInMainFile())"],
        "id": "Exception",
        "description": "Simplify your error handling - you can achieve this in other ways."
      }
    ]
  },

  "chapter_03_data": {
   ...
}
```
Each rule set is named (e.g "chapter_04_control_flow"), and contains a list of rules. Each rule contain a list of queries to try for that rule, and an ID and description/message that can be output when that rule is matched. Each rule set can also include the rules from another rule set.

## Type of change

- [?] This change requires a documentation update

# How Has This Been Tested?

A test script has been included (`concept_usage_test.py`) that tests a set of representative (but shortened/slightly mangled) pieces of code from the Field Guide. It checks that they pass under the rule sets that are from their own chapter or later, and fail with rule sets from earlier chapters. It also checks the specific `id`s of the rule matches to ensure they are matching correctly.

That's not to say this is exhaustive - I found a bunch of edge cases while putting the tests together, so it is possible there are more :sweat_smile: 

# Checklist:

- [ ] My code follows the style guidelines of this project (it's in Python sorry, hopefully that's okay...)

### DIFF
--- a/app/models/concept_usage/check_concept_usage.py
+++ b/app/models/concept_usage/check_concept_usage.py
@@ -1,0 +1,188 @@
+#!/usr/bin/env python3
+
+import json
+import subprocess
+import sys
+from dataclasses import dataclass
+from typing import Dict, List, Set
+import re
+import traceback
+import os.path
+
+# ==================== Constants ======================
+RULES_PATH = "rules.json"
+
+# Return codes
+SUCCESS = 0
+CODE_HAD_ISSUES = 32
+CODE_CHECKING_FAILED = 64
+
+# Removes all output except "matched" messages
+# No user code should be visible in the output
+CLANG_QUERY_PREAMBLE = """\
+set traversal IgnoreUnlessSpelledInSource
+disable output diag
+disable output print
+disable output detailed-ast
+disable output dump
+"""
+
+# ==================== Clang Query Utils ======================
+
+# Run clang-query and return the result
+
+# Note: it deliberately ignores any stderr output,
+# since we can't guarantee student's code will parse perfectly
+# on our system (they could have their own include file, etc).
+# Clang does a best effort parse in these cases, seems acceptable.
+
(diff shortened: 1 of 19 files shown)


b0ink:
~~Did you open this PR in the correct repository?~~

b0ink:
Added to the [SplashKit Docker Image](https://github.com/b0ink/splashkit-docker-image) so that it can be used in Overseer workflows

WhyPenguins:
@b0ink Ah my apologies, indeed I intended to close this since it was moved to [splashkit/cpp-language-feature-check](https://github.com/splashkit/cpp-language-feature-check/). Looks like you already have that sorted though, thanks for adding it :)

PR #607, opened 2026-04-09
feat: capture daily snapshots of aggregated task completion data

audreypho:
# Description

Adds daily scheduled snapshots of task completion data for each unit aggregated at a tutorial level by task and status.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] units_api_test.rb
- [ ] task_completion_snapshot_test.rb
- [x] unit_model_test.rb

# Checklist:

If you have any questions, please contact @macite or @jakerenzella.


app/models/unit.rb
```diff
@@ -3116,27 +3108,63 @@ def get_tutor_times_csv(start_date: nil, end_date: nil, timezone: nil, ignore_se
...
+      result[campus_name] ||= {}
+
+      task_definitions.each do |td|
+        tutorial = project.tutorial_for(td)
+        tutorial_name = tutorial&.abbreviation || 'Unassigned'
+
+        result[campus_name][tutorial_name] ||= {}
+        result[campus_name][tutorial_name][td.abbreviation] ||= {}
+
+        task = project.task_for_task_definition(td)
```
b0ink, line 3146: Use `has_task_for_task_definition?` instead, or even `status_for_task_definition`, as these won't create an empty task if the student hasn't submitted anything yet and default to `not_started`

Make sure you're still counting +1 for not_started if the task doesnt exist

db/migrate/20260322143502_create_task_completion_snapshots.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateTaskCompletionSnapshots < ActiveRecord::Migration[8.0]
+  def change
+    create_table :task_completion_snapshots do |t|
+      t.references :unit, null: false, foreign_key: true
```
b0ink, line 6: Remove the `foreign_key: true` attribute and re-run the migration, we typically avoid these because we define relationships via `belongs_to :unit` within the model, all handled by ActiveRecord

db/migrate/20260322143502_create_task_completion_snapshots.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.date :snapshot_date, null: false
+      t.datetime :captured_at, null: false
+      t.json :stats, null: false
```
b0ink, line 9: Whether we store the data as a JSON file or in the database may need a discussion, the data is probably minimal but if we store them as JSON files, we'll be able to archive them into a zip file at the end of the semester

We'd probably keep this model, but do something similar to the `OverseerAssessment` where we track a `submission_timestamp`, and create a directory/file named after that timestamp, so we'd be able query this table, and then locate the JSON snapshot in something like `student-work/{unit}/analytics/task-statuses/{timestamp}.json`. This way the JSON is easily downloadable and can all be archived/compressed at the end of the semester, without having to delete any  TaskCompletionSnapshots from the database

app/api/units_api.rb
```diff
@@ -574,6 +574,63 @@ class UnitsApi < Grape::API
...
+      {
+        snapshot_date: snapshot.snapshot_date,
+        captured_at: snapshot.captured_at,
+        stats: converted_stats
+      }
+    }, with: Grape::Presenters::Presenter
+  end
+
+  desc 'Capture task completion snapshot immediately for this unit'
+  post '/units/:id/stats/task_completion_snapshots/capture' do
```
b0ink, line 615: `AggregateTaskCompletionStatsJob.perform_async` sidekiq job should be used instead since this process may take up a couple minutes in production, you can see an example [here](https://github.com/doubtfire-lms/doubtfire-api/blob/f41d2ae6741a02552fee17d801ddefbfe99e7c75/app/api/units_api.rb#L518-L520) of starting the job and returning the job id

You can then use the [sidekiqProgressModalService](https://github.com/doubtfire-lms/doubtfire-web/blob/c47ded6d7b1c355e7c2efad7b9cb091bc0753241/src/app/units/states/analytics/unit-analytics-route.component.ts#L56-L81) to display the progress of the job and refresh the data in the final callback when viewing the result when the job has completed

app/api/units_api.rb
```diff
@@ -574,6 +574,63 @@ class UnitsApi < Grape::API
...
+    unit = Unit.find(params[:id])
+    unless authorise? current_user, unit, :download_stats
+      error!({ error: "Not authorised to capture stats of student tasks in #{unit.code}" }, 403)
+    end
+
```
b0ink, line 620: We should also rate limit this call so that if a snapshot exists within the past 30? 60 minutes? that it can't be called again

app/api/units_api.rb
```diff
@@ -574,6 +574,63 @@ class UnitsApi < Grape::API
...
+        captured_at: snapshot.captured_at,
+        stats: converted_stats
+      }
+    }, with: Grape::Presenters::Presenter
+  end
+
+  desc 'Capture task completion snapshot immediately for this unit'
+  post '/units/:id/stats/task_completion_snapshots/capture' do
+    unit = Unit.find(params[:id])
+    unless authorise? current_user, unit, :download_stats
```
b0ink, line 617: Restrict the snapshot capture to convenors only, so we could create a new permission for the `convenor_role_permissions` in `unit.rb`

b0ink (commented):
Looking good - we will need to discuss whether we store the aggregated data inside the database or if they should be saved as json files, mainly for easy archiving/compression at the end of the semester

Can you also fix the merge conflicts so i can run the tests? - I recommend to just rollback the migration via `rails db:rollback`, pull the latest schema changes, then start a new migration so that you have a newer migration timestamp, then re-run the migration with the requested changes (removing foreign key)

db/migrate/20260322143502_create_task_completion_snapshots.rb
audreypho: Completed but now in process of changing this to CSV or JSON storing all data available in task completion stats instead of the aggregated data.

app/models/unit.rb
```diff
@@ -1803,7 +1805,7 @@ def task_completion_csv
       # Get the details to fetch for each task definition...
       td_select = task_def_by_grade.map do |td|
         result = []
-        result << "MAX(CASE WHEN tasks.task_definition_id = #{td.id} THEN (CASE WHEN task_statuses.name IS NULL THEN 'Not Started' ELSE task_statuses.name END) ELSE NULL END) AS status_#{td.id}"
+        result << "MAX(CASE WHEN tasks.task_definition_id = #{td.id} THEN (CASE WHEN tasks.task_status_id IS NULL THEN #{TaskStatus.not_started.id} ELSE tasks.task_status_id END) ELSE NULL END) AS status_#{td.id}"
```
b0ink, line 1808: ~~Revert these back to the status names for clarity, but instead of `'Not Started'`, we can leave this as an empty string `... task_statuses.name IS NULL THEN '' ELSE task_statuses.name ...` to improve on file size~~

~~(Once we archive all the snapshots the task status names will compress down)~~

In the backend we can store these as the status IDs, then when a tutor wants to download the CSV, we rebuild the CSV to map them to the task status names - ensure tests are updated to match this change

app/models/unit.rb
```diff
@@ -1854,7 +1856,7 @@ def task_completion_csv
           end.flatten + grp_sets.map do |gs|
             row["grp_#{gs.id}"]
           end + task_def_by_grade.map do |td|
-            result = [row["status_#{td.id}"].nil? ? TaskStatus.not_started.name : row["status_#{td.id}"]]
+            result = [row["status_#{td.id}"].nil? ? TaskStatus.not_started.id : row["status_#{td.id}"].to_i]
```
b0ink, line 1859: Same here, revert to the task status name

app/models/unit.rb
```diff
@@ -3450,6 +3452,107 @@ def get_tutor_times_csv(start_date: nil, end_date: nil, timezone: nil, ignore_se
     end
   end
 
+  def aggregate_task_complete_stats
```
b0ink, line 3455: Is `aggregate_task_complete_stats` essentially the same as `task_completion_csv`? Can this be removed since we just need to call `task_completion_csv` and store that snapshot?

b0ink, line 3455: Can we also bring back the campus filtering? You'll need to modify the query to fetch the project's campus and add it as a new column to the CSV in `task_completion_csv` so that we can that split the task statuses even further by Campus -> Tutorial -> Tasks in the front end

app/models/unit.rb
b0ink, line 1808: Another thing we can consider is to always store these snapshots in a ZIP file. Rails is good at reading and extracting files from a ZIP file on the fly

14 weeks of daily snapshots at around ~700KB, we might be looking at around ~70MB worth of snapshots per unit. Since there's gonna be a lot of repeated text across the CSV snapshots, the file size will be minimal once compressed

b0ink (commented):
Overall workflow is looking really good

Just a few notes about bringing back the task status names, and querying the student's campus so we can bring back the campus filtering in the front end

Can you also please fix the RuboCop errors in the gh actions?

app/models/unit.rb
b0ink: This change breaks other existing tests too, so it's probably best to just keep the `'Not Started'` text as well

app/helpers/file_helper.rb
```diff
@@ -276,6 +276,27 @@ def unit_portfolio_dir(unit, create: true, archived: true)
...
+    dst << 'task-statuses/'
+
+    FileUtils.mkdir_p(dst) if create
+    dst
+  end
+
+  def unit_task_status_snapshot_path(unit, snapshot_timestamp, create: true, archived: true)
+    snapshot_filename = "#{sanitized_filename(snapshot_timestamp.to_s)}.zip"
+    File.join(unit_task_status_snapshots_dir(unit, create: create, archived: archived), snapshot_filename)
+  end
```
b0ink: Rather than storing each snapshot in its own zip file, it would be more efficient to store a single `task-status-snapshots.zip` in `{unit}/analytics/task-status-snapshots.zip` where we then place all the CSV files inside.

You'll need to then search for the timestamped snapshot within the ZIP

app/models/unit.rb
```diff
@@ -1825,20 +1840,24 @@ def task_completion_csv
...
-          'projects.id', 'student_id', 'username', 'first_name', 'last_name', 'target_grade', 'email', 'compile_portfolio', 'portfolio_production_date', 'grade', 'grade_rationale'
+          'projects.id', 'student_id', 'username', 'first_name', 'last_name', 'campus_abbreviation', 'target_grade', 'email', 'compile_portfolio', 'portfolio_production_date', 'grade', 'grade_rationale'
         ).each do |row|
-          csv << ([
+          student_details = [
             row['student_id'],
             row['username'],
             "#{row['first_name']} #{row['last_name']}",
+          ]
+          student_details << row['campus_abbreviation'] if includes_campus
```
b0ink: You can simplify this by just always including `campus` in the CSV, so you can remove the `includes_campus` argument.

`task_status_uses_id` should be fine to stay

b0ink (approved):

